PR #146, opened 2023-07-19
modify pdp-sharding docs to copy-paste level

omer9564:


docs/concepts/pdp-sharding.mdx
```diff
@@ -44,110 +44,222 @@ Permit.io's backend uses that very same sharding algorithm, Thus both the locall
...
+                        socket_address:
+                          address: pdp-shard-6
+                          port_value: 7000
+                    metadata:
+                      filter_metadata:
+                        envoy.lb:
+                          hash_key: "6"
    ```
 
+2. Perform HTTP request to notify a migration to sharded PDP cluster is being made ( this request must be done with env level api-key ), the response json will contain a `client_secret` that should be used in the next steps as the new cluster PDP_API_KEY:
```
asafc, line 160: what is this doing? will the previous non-sharded PDP be able to still run until the move to shards?

asafc (approved):
looks fine for now.
is this deployed to all environments or only uswest?

docs/concepts/pdp-sharding.mdx
omer9564: yes it doesn't effect the old PDP, it only creates a new one ( with the same name in the same env )

PR #172, opened 2023-09-06
Omer/per 7811 new rules interfaces

omer9564:


docs/how-to/enforce-permissions/all-tenants-check.mdx
```diff
@@ -0,0 +1,117 @@
...
+Enforcing with relationships ( ReBAC ) is not supported for Any Tenant Check.
+This is because the relationships check is performed on a resource instance, which is always scoped to a single Tenant.
+The output will always be an empty array if it isn't allowed in the scoped Tenant and
+an array with a single element if it is allowed, and therefore it is better to use simple check for better performance.
+
+:::
+
+:::warning
+
+When using [PDP Sharding](/concepts/pdp-sharding), the Any Tenant Check operation will not work properly.
```
filipermit, line 113: Please note that the "Any Tenant Check" feature might not work correctly if you're using [PDP Sharding](/concepts/pdp-sharding). This is because it only runs on one PDP instance at a time, and that instance only contains data for a few tenants, not all of them in the setup.

docs/how-to/enforce-permissions/bulk-check.mdx
```diff
@@ -0,0 +1,71 @@
+---
+sidebar_position: 2
+title: Bulk Check
+---
+
+
+To perform bulk permission checks, the `permit.BulkCheck` function can be leveraged.
+This allows you to validate multiple permission requests in a single call.
+
+## Usage
```
filipermit, line 10: Basic Usage*

docs/how-to/enforce-permissions/bulk-check.mdx
```diff
@@ -0,0 +1,71 @@
...
+				)
+			}
+		}
+	}
+}
+```
+
+:::note
+
+When using [PDP Sharding](/concepts/pdp-sharding), the Bulk Check operation will only work for check requests in the same tenant.
```
filipermit, line 67: Same as above ;)

filipermit (approved):

PR #207, opened 2023-11-16
PER-8084 add relationship tuples filter page

noamovich19:


docs/api/examples/filter-relationship-tuple.mdx
```diff
@@ -0,0 +1,112 @@
...
+            "description": null,
+            "attributes": null
+        }
+    }
+]
+```
+
+## Filter
+filter the relationship tuples data
+:::note Note you can mix and match the filters
```
omer9564, line 65: You need to close the "note" with :::
```suggestion
:::note Note you can mix and match the filters
this is equivalent to logical `AND` between them
:::
```

omer9564: Also I would put it below and not above but that is only my preference

omer9564 (changes requested):

PR #263, opened 2024-03-04
api pdp docs

EliMoshkovich:


RazcoDev:
It looks a bit weird with the openapi.json, there's no really need in it-  you can remove it: )
Also, you can remove the "Permit API object hierarchy" part.

RazcoDev:
Also, in order to not break the url - let's keep the cloud api page in the same name, just change the title of it. 
Think about that we are using this page in other places and now the page URL will be broken. 
It's not the best practice - the better way to handle this is creating a redirect.
I'm ok with this quick fix right now because we are planning to change this page soon in our docs upgrade effort.

RazcoDev (approved):

PR #277, opened 2024-03-07
Add docs to unauthenticated env var

obsd:


RazcoDev:
Set this as an admonition (`:::info`) so it'll be noticeable :D

RazcoDev (approved):

PR #300, opened 2024-03-29
Document bulk role route

orishavit:


docs/how-to/bulk-operations.mdx
```diff
@@ -189,3 +189,27 @@ curl -X DELETE https://api.permit.io/v2/facts/{proj_id}/{env_id}/relationship_tu
             }, ...]
         }'
 ```
+## Bulk role operations
+### Create roles
+
+Create multiple roles in one request. You may create regular roles, as well as ReBAC resource roles for multiple resource in the same request. Pass the optional "resource" field to create a resource role. Roles created without the resource field will be created as regular (RBAC) roles. If the role exists, it will be updated.
```
omer9564, line 195: add a reference to the redoc

docs/how-to/bulk-operations.mdx
```diff
@@ -189,3 +189,27 @@ curl -X DELETE https://api.permit.io/v2/facts/{proj_id}/{env_id}/relationship_tu
...
+                    "key": "admin",
+                    "name": "Administrator",
+                    "permissions": ["document:create", "document:delete"],
+                },
+                {
+                    "key": "editor",
+                    "name": "Editor",
+                    "resource": "document"
+                    "permissions": ["read", "write"],
+                }...]
```
omer9564, line 213: I would remove the ... so it will be copy paste valid

omer9564 (changes requested):

omer9564 (approved):
LGTM

PR #314, opened 2024-04-07
Add docs for list role assignments from the PDP

omer9564:


asafc (approved):
LGTM, but maybe add an example output so that the format of the response is also documented?

omer9564:
@asafc Good idea, will do after I create the PDP schema and have an actual example

PR #338, opened 2024-05-05
add doc for the Default Roles and advanced policy settings

maya-barak:


docs/api/examples/manage-env.mdx
```diff
@@ -47,6 +47,39 @@ The return environment object will look like this:
 }
 ```
 
+
+:::note
+When creating a new environment from the dashboard, if the organization settings "enable_default_roles" is set to true,
+the environment settings will be configured to the default values.
```
RazcoDev, line 53: That's a huge note - doesn't look so good. 
Better it to be a sub title "Environment's Default Permissions"

RazcoDev (approved):

PR #372, opened 2024-06-21
New Groups API version for review

danielbass37:
Still missing diagram. 
Might have some mistakes and require fixing the API code. 
Please review.

### DIFF
--- a/docs/api/rebac/groups.mdx
+++ b/docs/api/rebac/groups.mdx
@@ -6,60 +6,86 @@ title: Groups API
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
-The Groups API allows you to manage your resources and users within groups. You can create, update, delete, and list groups.
-Groups enable you to easily group together resources, resource-instances, and users. In addition groups simplify ReBAC permissions for users, by allowing roles to be assigned to the groups themselves.
-You can assign roles to a group and then add users to the group through the Groups API.
-The users within the group will derive roles form the roles assigned to the group .
+The Groups API simplifies permission management for your resources and users by using groups. Groups can include resources, resource instances, and users, enabling you to set permissions for everything contained within a group at once.
 
-Check the [Groups Redoc ](/) for more information.
+## Key Features:
 
+- **Create, Update, Delete, and List Groups:** The Groups API provides a comprehensive set of endpoints to manage the lifecycle of your groups, ensuring you have complete control over their configuration and membership.
+- **User Membership:** When adding users to the group through the API, you can set them to automatically derive their permissions from the roles assigned to the group, ensuring consistent and manageable access control. See the [Groups Redoc](https://www.notion.so/) for more information.
+- **Group Resources, Resource Instances, and Users:** Organize your resources and users into groups, allowing for simplified ReBAC permissions for:
+
+    - **Users:** Creating a group of users allows you to assign an entire cluster of users with a single role (Instead of assigning roles to each user individually). You can set a role assigned to the group to be automatically assigned to all its members and revoked once their membership is removed. 
+    
+    :::info Example:
+    If we create a `marketing` group and assign a `marketing-team-member` role to it, we can give an entire group of users access to a shared marketing resource.
+    :::
+    
+    - **Resources:** Grouping resources enables you to manage access control for multiple resources simultaneously, ensuring all included resources share the same permission configuration.
+    
+    :::info Example:
+    If we create a `marketing-resources` group, we can assign a role to a user (Or group of users!) that gives them permission to perform certain actions on all resources within the `marketing-resources` group.
+    :::
+
+    - **Resource Instances:** Grouping resource instances allows you to manage permissions for a group of specific resource instances collectively. As with resource groups, this ensures all resource instances within the group share the same permission configuration.
+        
+    :::info Example: 
+    If we create a `social-media-campaigns` group and populate it with the instances of our social media campaigns (`campaign-1`, `campaign-2`), we can assign a user (Or group of users) a role that gives them access to perform specific actions on this group of resource instances. 
+    :::
+        
 
 ## Example Usage for User Groups
 
-### Assign roles to a group:
+Let's say we want to assign a group of users, the "Marketing Team," access to a group of social media resources. 
 
-Let's say we want to represent a company and its employees as a group, and then assign all the employees access to a shared marketing resource (social-media) via the group.  We'll have a group named 'company' (Which is represented as a resource named 'company'), and an instance of a specific company 'Acme'.  Users are assigned to `Acme` via a role called `member`
-Additionally, we will have a resource-instance of type 'marketing' - 'social_media' which supports the role 'editor'.
-Once we assign the role 'editor' to  the `company` group, the following will by automated by API:
-- create role derivation from 'company#member' to 'marketing#editor'
-- create a relation between 'company' and 'marketing' with a relation between 'company_group'
-- create relationship between 'company#Acme' and 'marketing#social_media'.
+- Using the Groups API, we'll first create a group named `marketing-team`.
+- Next, we will assign users to the `marketing-team` group.
+- The group will be assigned a `member` role.
 
-![](groups-diagram.png)
+- Additionally, we will have a resource group called `social-media` .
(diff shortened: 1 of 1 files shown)


docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,86 @@ title: Groups API
...
-Groups enable you to easily group together resources, resource-instances, and users. In addition groups simplify ReBAC permissions for users, by allowing roles to be assigned to the groups themselves.
-You can assign roles to a group and then add users to the group through the Groups API.
-The users within the group will derive roles form the roles assigned to the group .
+The Groups API simplifies permission management for your resources and users by using groups. Groups can include resources, resource instances, and users, enabling you to set permissions for everything contained within a group at once.
+
+## Key Features:
+
+- **Create, Update, Delete, and List Groups:** The Groups API provides a comprehensive set of endpoints to manage the lifecycle of your groups, ensuring you have complete control over their configuration and membership.
+- **User Membership:** When adding users to the group through the API, you can set them to automatically derive their permissions from the roles assigned to the group, ensuring consistent and manageable access control. See the [Groups Redoc](https://www.notion.so/) for more information.
```
maya-barak, line 14: group redoc- https://api.permit.io/v2/redoc#tag/Groups-(Upcoming-Feature)

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,86 @@ title: Groups API
...
+- **Group Resources, Resource Instances, and Users:** Organize your resources and users into groups, allowing for simplified ReBAC permissions for:
+
+    - **Users:** Creating a group of users allows you to assign an entire cluster of users with a single role (Instead of assigning roles to each user individually). You can set a role assigned to the group to be automatically assigned to all its members and revoked once their membership is removed. 
+    
+    :::info Example:
+    If we create a `marketing` group and assign a `marketing-team-member` role to it, we can give an entire group of users access to a shared marketing resource.
```
maya-barak, line 20: the 'marketing-team-member' role and 'marketing-resources' are not expressions that fit for what we wanted for this example, @orweis - your opinion?

docs/api/rebac/groups.mdx
orweis: Please don't have "Upcoming feature" in the URL-  that means the URL will break once that changes.
Use admonitions in the text instead

docs/api/rebac/groups.mdx
orweis: Let's switch the resource to be (an external) `document`

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,86 @@ title: Groups API
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
-The Groups API allows you to manage your resources and users within groups. You can create, update, delete, and list groups.
-Groups enable you to easily group together resources, resource-instances, and users. In addition groups simplify ReBAC permissions for users, by allowing roles to be assigned to the groups themselves.
-You can assign roles to a group and then add users to the group through the Groups API.
-The users within the group will derive roles form the roles assigned to the group .
+The Groups API simplifies permission management for your resources and users by using groups. Groups can include resources, resource instances, and users, enabling you to set permissions for everything contained within a group at once.
```
GonPessach, line 9: "Groups can include resources, resource instances..." - you can group resource types to set up the scheme of your app resources, later the permission will be managed on resource instances groups according the the scheme that was set on the type. so when saying "Groups can include resources, resource instances and users..." I think it communicates the wrong idea. 

" enabling you to set permissions for everything contained within a group at once" - only relevant for user groups. I think it's confusing in the context of resource groups.

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,86 @@ title: Groups API
...
+    If we create a `marketing` group and assign a `marketing-team-member` role to it, we can give an entire group of users access to a shared marketing resource.
+    :::
+    
+    - **Resources:** Grouping resources enables you to manage access control for multiple resources simultaneously, ensuring all included resources share the same permission configuration.
+    
+    :::info Example:
+    If we create a `marketing-resources` group, we can assign a role to a user (Or group of users!) that gives them permission to perform certain actions on all resources within the `marketing-resources` group.
+    :::
+
+    - **Resource Instances:** Grouping resource instances allows you to manage permissions for a group of specific resource instances collectively. As with resource groups, this ensures all resource instances within the group share the same permission configuration.
```
GonPessach, line 29: continuing the comment from above, I think there is some misunderstanding around the resource group use case, would love to chat about this more

docs/api/rebac/groups.mdx
danielbass37: @maya-barak please fix the link before I push changes.

docs/api/rebac/groups.mdx
danielbass37: For the first note: I agree. Changed to: "These can be used to group together resources, resource instances, or users,"

As for the second note, how's: "enabling you to set a unified permission scheme for everything contained within a group"? I need something that works for all three group types..

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,93 @@ title: Groups API
...
-Once we assign the role 'editor' to  the `company` group, the following will by automated by API:
-- create role derivation from 'company#member' to 'marketing#editor'
-- create a relation between 'company' and 'marketing' with a relation between 'company_group'
-- create relationship between 'company#Acme' and 'marketing#social_media'.
 
-![](groups-diagram.png)
+- **Group Resource Instances, and Users**: Organize your resource instances and users into groups, allowing for simplified ReBAC permissions for:
 
-Then we will have group 'company' and every user who is a 'member' in this group will have the role 'editor' in the resource 'marketing#social_media'.
+    - **Users:** Users: Including users in a group allows you to assign an entire cluster of users with a single role (Instead of assigning roles to each user individually).
```
orweis, line 22:
```suggestion
    - **Users:** Including users in a group allows you to assign an entire cluster of users with a single role (Instead of assigning roles to each user individually). 
```

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,93 @@ title: Groups API
...
-- create relationship between 'company#Acme' and 'marketing#social_media'.
 
-![](groups-diagram.png)
+- **Group Resource Instances, and Users**: Organize your resource instances and users into groups, allowing for simplified ReBAC permissions for:
 
-Then we will have group 'company' and every user who is a 'member' in this group will have the role 'editor' in the resource 'marketing#social_media'.
+    - **Users:** Users: Including users in a group allows you to assign an entire cluster of users with a single role (Instead of assigning roles to each user individually). 
+    You can set a role assigned to the group to be automatically assigned to all its members and revoked once their membership is removed.
+    
+    :::info Example:
```
orweis, line 25: Hoe is this an example ?

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,93 @@ title: Groups API
...
+    :::
+    
+    - **Resource Instances**: Grouping resources enables you to manage access control for multiple resources simultaneously, ensuring all included resources share the same permission configuration.
 
-### Assign users to a group:
+    :::info Example:
+    In our marketing group, we can create another group titled `social_media`. This group will contain resource instances of marketing materials. We can assign this newly created group the role of editor. 
+    The Groups API allows you to create a derivation between the `social_media#editor` role and the `marketing#member` role - meaning all users with the `marketing#member` role will automatically be assigned 
+    with the `social_media#editor` role.
+    :::
```
orweis, line 35: You're confusing giving an example of usage, and what would be created behind the scenes  

The groups API doesn't "allows you to create a derivation" it does so for us automatically

docs/api/rebac/groups.mdx
```diff
@@ -6,60 +6,93 @@ title: Groups API
...
+    :::info Example:
+    New resources added to `social_media` will be accessible to edit by members of the marketing group, and new users added as members to the marketing group will automatically 
+    have editor access to social media assets.
+    :::
```
orweis, line 39: This is not a new example - it's part of the previous one.
Not sure this needs to be an admonition at all

PR #381, opened 2024-07-03
add timeout doc

maya-barak:


docs/how-to/enforce-permissions/bulk-check.mdx
```diff
@@ -162,6 +162,18 @@ const permittedResources = resources.filter(
 ```
 
 ---
+### Set Timeout
+For setting the timeout for the bulk check, you can set the 'PDP_OPA_CLIENT_QUERY_TIMEOUT' environment variable for the PDP.
```
obsd, line 166: Bulk check is more performance intensive than regular check and also takes more time which can lead to OPA timeout.
To address this issue, you can increase the query timeout for the PDP and add more CPU to your VM.

docs/how-to/enforce-permissions/bulk-check.mdx
```diff
@@ -162,6 +162,18 @@ const permittedResources = resources.filter(
...
+
+If for example you want a 10 second timeout, you can set the environment variable as follows:
+
+```bash
+PDP_OPA_CLIENT_QUERY_TIMEOUT=10
+```
+OPA Runs the authorization query concurrently based on the amount of CPU it has, so if you have a lot of queries to run, you can increase the CPU count to speed up the process.
```
obsd, line 173: Are you sure about that?
```
OPA Runs the authorization query concurrently based on the amount of CPU it has
```

orweis:
This is good, but I think you misunderstood my ask- I asked for docs for all the PDP config variables

PR #386, opened 2024-07-14
dan/per-10258-add-docs-on-all-the-pdp-config

danyi1212:


docs/concepts/pdp/configuration.mdx
```diff
@@ -0,0 +1,346 @@
...
+#### OPAL_CLIENT_TOKEN
+
+Default: automatically configured
+
+The client token used to authenticate the PDP against the OPAL server.
+Automatically configured by the control plane.
+
+#### OPAL_INLINE_OPA_ENABLED
+
+Default: `True`
```
omer9564, line 269: I'm not sure about the purpose of this documentations as you can't really configure them ( and even if you can you shouldn't because it will break things )
@obsd @orweis WDYT ?

docs/concepts/pdp/configuration.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+The log format for OPA logs.
+* `none` - no logs
+* `minimal` - only the event name is logged
+* `http` - log HTTP method, path, and status code
+* `full` - log the entire data returned
+
+#### OPAL_INLINE_CEDAR_ENABLED
+
+Default: `True`
```
omer9564, line 285: is this really default True ? doesn't seem right

docs/concepts/pdp/configuration.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+Default: `60`
+
+The timeout for fetching the callback, in seconds.
+
+#### OPAL_FETCHING_ENQUEUE_TIMEOUT
+
+Default: `10`
+
+The timeout for queueing a fetch, in seconds.
```
omer9564, line 315: these are the only configurations that I think that are relevant here

docs/concepts/pdp/overview.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+# Permit's PDP (Policy Decision Point)
+
+A PDP is a network node responsible for answering authorization queries using policies and contextual data.
+Hence, PDPs should be highly available, performant and physically close to the querying services to minimize latency and ensure high availability.
+
+Permit's PDP essentially becomes your microservice for authorization, bundling together OPA, OPAL and an API Server.
+
+The PDP is easy to install via Docker image, and will ensure **zero-latency, great performance, high availability, and improved security**.
+Like all of Permit's customer deployed components, the [Permit PDP is open-source](https://github.com/permitio/PDP) and is available publicly from [Docker hub](https://hub.docker.com/r/permitio/pdp-v2).
```
omer9564, line 14: isn't this duplicated ?

docs/concepts/pdp/overview.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+Permit.io supports all PDP layouts and provides the missing layers on-top of open-source PDP solutions (such as OpenPolicyAgent). These layers include policy delivery and updating, supporting data collection, application level SDKs, application level instrumentation and more.
+Read about [the various layouts you can deploy the PDP with here](/how-to/deploy/overview#pdp-deployments).
+
+There are several ways you can deploy the PDP:
+
+
+### Managed Cloud PDP
+
+Permit.io provides a managed Cloud PDP at `https://cloudpdp.api.permit.io` which is available globally.
```
omer9564, line 26: you need to add important note that this "cloud pdp" only supports RBAC ( doesn't support ReBAC / ABAC )

docs/concepts/pdp/overview.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+
+### Hosted as a Sidecar
+
+The simplest way to integrate PDPs into a microservices architecture is as sidecars;
+meaning each microservice has a sidecar container next to it which it can query for policy.
+
+For more information, see the [sidecar deployment guide](/how-to/deploy/overview#sidecar).
+
+```mermaid
```
omer9564, line 133: in the graphs it looks like the pdp is interacting with the cloud, I think it is better to change the arrow to be from the cloud to the PDP to show that the pdp doesn't query the cloud but the cloud updates it

docs/concepts/pdp/overview.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+    style cloud fill:#ffb381,color:#000;
+    style app1 fill:#974ef2,color:#fff;
+    style app2 fill:#974ef2,color:#fff;
+    style pdp1 fill:#f76808,color:#fff;
+    style pdp2 fill:#f76808,color:#fff;
+    style envoy fill:#4e7ef2,color:#fff;
+```
+
+## Powered by OPA+OPAL
```
omer9564, line 233: same here, I think we would want to change the powered by OPA+OPAL as we will support more engines in the future.
@orweis @obsd WDYT ?

docs/concepts/pdp/configuration.mdx
danyi1212: Yes, its True in real life. You can run PDP and print the config to see its enabled.

I think it does not work because it just doesn't have the cedar agent built in the image

docs/concepts/pdp/overview.mdx
danyi1212: 90% of the docs here are duplicate🙈 I think it fits here

docs/concepts/pdp/overview.mdx
danyi1212: I tried doing it, but it messed up the chart. I've now found bidirectional arrows and I think it fits best

omer9564 (approved):

PR #398, opened 2024-08-06
add docs to the fk object match abac operator

omer9564:
the added docs can be viewed at /api/working-with-abac/operators#object-match-using-a-foreign-key

### DIFF
--- a/docs/api/working-with-abac/operators.mdx
+++ b/docs/api/working-with-abac/operators.mdx
@@ -230,4 +230,11 @@ This can also be done with `any_match`:
 }
 ```
 
+:::info
+While other ABAC operators can be used by attributes from the check arguments or attributes stored in Permit/PDP, <br />
+the foreign key condition aims to solve cases where we already store the attributes in Permit Cloud and
+want to have a condition that does not need to pass them explicitly as part of the check function.
+:::
+
+
 Now that we have a basic understanding of the language to construct conditions, lets jump into creating **[Condition Sets](condition-sets)**.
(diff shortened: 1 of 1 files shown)


docs/api/working-with-abac/operators.mdx
```diff
@@ -164,5 +164,70 @@ Finally if we want to check if all items in the array match a certain condition
...
 
+### Object Match using a "foreign key"
+
+In some cases, we have an attribute on one resource that references the key of other resource.<br/>
+For example, a resource `repo` has an attribute of `organization` ( string ) that holds the organization key of that repo,
+similar to a foreign key in a database.<br />
+In case we want to create a condition that checks if the organization of the repo is located in the US, we can the
+`object_match` operator with a reference to the organization object by adding the `fk_resource_type` key and provide the foreign resource type in it.<br />
+The following example will extract the **stored** attributes of the organization in the `organization` attribute of the `repo` resource queried
+and apply the conditions on those extracted organization attributes:
```
gemanor, line 175:
```suggestion
and apply the conditions on those extracted organization attributes:

:::info
While other ABAC operators can be used by attributes from the check arguments or attributes stored in Permit/PDP, the foreign key condition aims to solve cases where we already store the attributes in Permit/PDP and want to have a condition that does not need to pass them explicitly as part of the check function.
:::


```

gemanor (approved):
LGTM, added optional admonition for your consideration

docs/api/working-with-abac/operators.mdx
omer9564: added this info

PR #404, opened 2024-08-14
change role to top level

shuvyA:
https://linear.app/permit/issue/PER-10442/change-roles-and-resource-roles-to-top-level-and-instance-roles-docs


RazcoDev (approved):
looks good well done

PR #418, opened 2024-09-17
fix user key to be clear

maya-barak:


docs/how-to/bulk-operations.mdx
```diff
@@ -48,7 +48,7 @@ curl -X PUT https://api.permit.io/v2/facts/{proj_id}/{env_id}/bulk/users
     -H "Content-Type: application/json" \
     -d '{
             "operations": [{
-            "key": "user|892179821739812389327",
+            "key": "jane_doe",
             "email": "
```
RazcoDev, line 52: Let's also fix this - it has only one `"` - but let's also set an email there so it will be a full example

RazcoDev (approved):

PR #419, opened 2024-09-18
Guarding Policy API

maya-barak:


docs/api/guarding-policy-api.mdx
```diff
@@ -0,0 +1,97 @@
+---
+sidebar_position: 7
+title: Policy Guard API
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+The Policy Guard API allows you to manage policies at the organization level. Policy Guards help enforce centralized control over permissions and roles across specific projects included in the policy.
+For example, you can create  that enforces a specific set of permissions, as role admin can create documents, and role viewer can only read documents, across all projects included in the policy.
```
RazcoDev, line 10: You are missing a word/or more after "you can create...."

RazcoDev: The example missing some grammar and structure. 
a better way to demonstrate an example is telling the example, and then demonstrating it, like - 
We want to make sure only admins can create a document, and all other roles cannot, to achieve this we'll create a policy guard that....

docs/api/guarding-policy-api.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+Let's see how you can use the Policy Guard API for the example above.
+
+
+## Example API Usage
+
+
+### Create a New Policy Guard Scope
```
RazcoDev, line 18: Add some text to explain the action - 
"First, we'll create a new policy guard scope in order to....

docs/api/guarding-policy-api.mdx
```diff
@@ -0,0 +1,97 @@
...
+ key":   "policy_guard_acme",
+         "org_id": "c3b6f5d7-8b1e-4c6d-9e9f-8c9d6f8e0c8f",
+         "policy_guard_scope_details": [
+             {
+                 "env_id": "c3b6f5d7-8b1e-4c6d-9e9f-8c9d6f8e0c8f",
+                 "proj_id": "c3b6f5d7-8b1e-4c6d-9e9f-8c9d6f8e0c8f",
+             }]
+}'
+```
+### List all Policy Guard Scopes
```
RazcoDev, line 32: "Now, we'll want to list the policy guard scopes to see that it is created properly....

docs/api/guarding-policy-api.mdx
```diff
@@ -0,0 +1,97 @@
...
+```bash
+curl 'https://api.permit.io/v2/policy_guards' \
+-H 'authorization: Bearer API_SECRET_KEY'
+```
+
+### Get a Policy Guard Scope by ID
```
RazcoDev, line 38: And so on

PR #420, opened 2024-09-18
dan/per-10462-add-async-copy-environment-api

danyi1212:


docs/api/background-tasks.mdx
```diff
@@ -0,0 +1,106 @@
...
+        "key": "prod",
+        "name": "production"
+    }
+  }
+}'
+# Response: 202 Accepted
+```
+When triggering a Copy Environment, the API will return a `202 Accepted` response, indicating that the job has been accepted and is being processed in the background.
+
+The response contains info about the task, including the `task_id` that can be used to check the status of the task.
```
obsd, line 33: Explain here how to check status (or at least add link)

docs/api/background-tasks.mdx
```diff
@@ -0,0 +1,106 @@
...
+```bash
+curl --location 'https://api.permit.io/v2/projects/default/envs/staging/copy/async/{task_id}/result' \
+  -H 'authorization: Bearer API_SECRET_KEY'
+```
+
+```json
+{
+  "task_id": "string",
+  "status": "success",
+  "result": {
```
obsd, line 55: Is this how the old copy env API looks like?

docs/manage-your-account/creating-environments.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+In this example, we have a Production environment with a policy that has a `File` Resource and an `Editor` Role.
+We've copied the Production environment into a Development environment, where we've added a `Folder` Resource and an `Admin` Role.
+After testing the changes, we've merged the Development environment back into the Production environment.
+
+This way, you can manage changes to your policy like code changes in your CI/CD pipelines.
+Read more about [Policy Life Cycle with Permit](/how-to/SDLC/CI-CD).
+
+:::warning Notice
+This feature is in **EAP** stage, which means it's in early access and might have some limitations.
```
obsd, line 85: No need to push this EAP to every place with long warnings
Put one info on this and in the rest of the places EAP is enough

docs/manage-your-account/creating-environments.mdx
```diff
@@ -174,10 +174,10 @@ gitGraph
 ```
 
 In this example, we've added an `Admin` Role to the Production environment after copying it into the Development environment.
-When we merge the Development environment back into the Production environment, we have a conflict with the `Admin` Role.
+When we merge the development environment back into the production environment, we will have a conflict with the `Admin` role.
 
 When merging environments, you can choose to resolve conflicts by either:
-* **Fail**: (default) Fail the merge, rollback the entire merge operation and keep the existing object on the target environment.
+* **Fail**: (default) If the merge fails, the entire merge operation will be rolled back, and the existing object will remain in the target environment.
```
danyi1212, line 180: Incorrect terminology. In case of merge conflict - it causes it to fail and roll back.
@danielbass37

docs/api/background-tasks.mdx
```diff
@@ -71,11 +71,11 @@ curl --location 'https://api.permit.io/v2/projects/default/envs/staging/copy/asy
...
 ## Waiting for Background tasks
 
-You can also wait for the background task to complete by using the `wait` query parameter.
+You can also wait for the background task to be completed by using the `wait` query parameter.
 
-The wait parameter specify the maximum time to wait for the task to complete in seconds.
+The `wait` parameter specifies the maximum time to wait for the task to complete in seconds.
 By default, `wait=0` will return immediately with the current status of the task.
-As soon as the task complete, the API will respond the result of the task.
+As soon as the task is complete, the API will respond to the result of the task.
```
danyi1212, line 78: Also became weird sentence. When the task is completed, the API will respond immediately with the result of the task.

docs/api/background-tasks.mdx
danyi1212: There is an example how to check one line after

docs/api/background-tasks.mdx
danyi1212: Yes, its the same type

obsd (approved):

PR #435, opened 2024-10-10
Added info about backoff retry for decision logs upload

danyi1212:


docs/concepts/pdp/configuration.mdx
```diff
@@ -230,6 +230,10 @@ Default: `1`
 
 The minimum amount of time (in seconds) to wait between decision log uploads.
 
+When unable to upload logs, the PDP performs an exponential backoff up to `PDP_OPA_DECISION_LOG_MAX_DELAY` until a successful upload.
```
omer9564, line 233: Not sure it is "until" a successful upload, from opa docs:
> This best-effort approach ensures that OPA reports decision log events as much as possible

Sounds like there is a scenario in which decision logs are dropped, am I wrong ? [ref](https://www.openpolicyagent.org/docs/latest/management-decision-logs/#decision-log-service-api)

omer9564 (approved):
I approve because I'm not sure about my comment, if you verified it than it's fine by me

docs/concepts/pdp/configuration.mdx
danyi1212: There technically has to be, otherwise it will compromise on new decisions, but it truly will still attempt forever to upload from that queue...

PR #446, opened 2024-11-10
FactDB docs

omer9564:
We used to not need to add the pages also to the sidebar.js because it was autogenerated or something like that.
I see that this isn't the case anymore ( or at least not the case here ).
Is this a known thing ? @gemanor

### DIFF
--- a/docs/concepts/pdp/factdb.mdx
+++ b/docs/concepts/pdp/factdb.mdx
@@ -37,7 +37,7 @@ curl --location --request PATCH 'https://api.permit.io/v2/projects/{proj_id}/env
 --header 'Authorization: Bearer {api_key}' \
 --data-raw '{
   "settings": {
-        "enable_external_data_store":true,
+        "factdb_enabled":true,
         "allow_legacy_data_store":true
     }
 }'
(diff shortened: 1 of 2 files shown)


gemanor:
@omer9564 it is a known thing, I shared about it on Slack. We did it to avoid extra redirects when moving pages in the docs and to help with better consistency in reviews.

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,55 @@
...
+---
+
+# FactDB ( EAP )
+
+FactDB is a new feature that allows you to store and query facts more efficiently in the PDP.
+It uses embedded DB technology to store facts in a more efficient way instead of the policy engine memory,
+and provides a RestAPI to query those facts.
+
+:::info EAP
+This feature is in EAP stage - please contact with us if you want to try it out.
```
obsd, line 13: I would edit this note
eap - subject to breaking changes
opt in feature - you should ask to join

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,55 @@
...
+:::
+:::warning
+To use this feature all your PDPs must be upgraded to version 0.7.0 or above.
+Otherwise, some features might not work as expected.
```
obsd, line 17: once enabled old pdps coneted to this env might have errors

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,55 @@
...
+:::
+
+## Enabling FactDB
+
+To enable the FactDB on your environment there are two configurations that you need to set:
```
obsd, line 22: 3, update pdp, setttings and env in the PDP
I would write headlines and only then have the instructions

danyi1212:
Will be merged on https://github.com/permitio/docs/pull/451

PR #451, opened 2024-11-27
dan/per-10914-docs-how-to-debug-data-with-the-new-data-sync

danyi1212:


omer9564 (approved):

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
+---
+sidebar_position: 11
+title: Local Datastore (FactDB)
+---
+
+# PDP Local Datastore (FactDB)
+
+FactDB is an embedded datastore to manage larges data volumes inside the PDP.
```
gemanor, line 8: Add link on the PDP word

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
+---
+sidebar_position: 11
+title: Local Datastore (FactDB)
+---
+
+# PDP Local Datastore (FactDB)
```
gemanor, line 6: 1. You don't need the #1 here, it will take it from the `title:` in the header.
2. Please convert the next 2 paragraphs to one TL;DR text that explains what is the outcome of reading the page, and not the objective of the page.
3. The TL;DR should be designed as the header of the page (not title, header)

See structure example here: https://github.com/permitio/docs/blob/master/docs/modeling/mesa-verde.mdx

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+It allows the PDP to store and query large volumes of data locally,
+reducing the latency and improving the performance of the PDP.
+
+:::info EAP
+This feature is available under Early Access Program.
+Please contact with us on [Slack](https://io.permit.io/slack) for more information.
+:::
+
+:::warning
```
gemanor, line 18: 1. Add a title keyword after the warning, so it will replace the `DANGER` default title.
2. It is also not the best practice to include admonitions one after the other. Better if you can find a better placement for that. You can move it after the first paragraph in the next section.

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+This version is available on **PDP v0.7.0** and above.
+Please make sure to upgrade **all PDPs in the environment** to the latest version to use FactDB.
```
gemanor, line 20: What is `the environment`? The Permit environment? The k8s cluster of the user?

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+:::
+
+## Enabling FactDB
+
+The FactDB can be enabled per Permit Environment, allowing PDPs in that environment to use the FactDB.
+Optionally, you can have both legacy PDPs and FactDB enabled PDPs in the same environment.
```
gemanor, line 26: This is a confusing way to state that. Should I enable it for the whole environment and then enable others? Why should I enable that?

I would recommend the following outline:
1. By default, Permit environment comes with FactsDB disabled.
2. You can enable the FactsDB in the environment level to support PDPs to run this way
3. When deploying PDP, you can decide if you want them to have local facts enabled.

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+  -H 'Content-Type: application/json' -H 'Accept: application/json' \
+  -H 'Authorization : Bearer {api_key}' \
+  --data-raw '{
+      "settings": {
+            "factdb_enabled":true
+      }
+    }'
+```
+:::info
+If you have existing settings in your environment, you may want to include them in the PATCH request.
```
gemanor, line 45: Isn't patch should let you patch only the relevant one? If we have a bug that requires you to send the old settings, we need to state it differently. (it's not you may want, it's you need...)

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+```bash
+docker run -it -p 7766:7000 \
+  --name pdp-online \
+  --env PDP_API_KEY=<YOUR_API_KEY> \
+  --env PDP_FACTDB_ENABLED=true \
+  permitio/pdp-v2:latest
+```
+
+:::info
+FactDB is currently an opt-in feature, requiring you to explicitly enable it on all PDPs in the environment.
```
gemanor, line 67: Maybe `every PDP` is more fit to the text?

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+This is to allow environments to run both with FactDB and without it.
```
gemanor, line 68: If you're enabling it in all the PDPs, how you can run it both with and without?

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+      "settings": {
+            "factdb_enabled":true,
+            "allow_legacy_data_store":true
+      }
+    }'
+```
+
+With this setting enabled, you can run both FactDB PDPs with `PDP_FACTDB_ENABLED=true` and Legacy PDPs in the same environment.
+
+:::info
```
gemanor, line 90: The overall structure of this section is very confusing. Should I enable in all PDPs? Can I use part of the PDPs? Are you recommending migrate gradually or not?

I think the outline should be more clear to the point.

1. Enable the Facts DB in the environment
2. If you'd like to do it only for some PDPs, add this configuration
3. When you run PDP, add this variable to run it with FactsDB

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+  --env PDP_FACTDB_ENABLED=true \
+  --env PDP_ENABLE_OFFLINE_MODE=true \
+  -v pdp-offline-backup:/app/backup:rw \
+  permitio/pdp-v2:latest
+```
+
+With the backup volume attached, even when the PDP is unable to connect to the Permit API,
+it can still boot up with the FactDB, policies and configurations from the previous run.
+
+```bash
```
gemanor, line 122: Please add a short explanation before like: Let's try to run the container with no network and see how's the offline mode load the FactsDB

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+  --env PDP_API_KEY=<YOUR_API_KEY> \
+  --env PDP_FACTDB_ENABLED=true \
+  --env PDP_ENABLE_OFFLINE_MODE=true \
+  -v pdp-offline-backup:/app/backup:rw \
+  permitio/pdp-v2:latest
+```
+
+## FactDB Architecture
+
+When the PDP boots, FactDB downloads the latest snapshot, and keep it up-to-date from the Permit Cloud and other PDPs in the environment.
```
gemanor, line 135: If you're using the `other PDPs...` in here, you should say When a PDP boots (or explain which is **THE** PDP

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+    participant opa as OPA
+    cloud  ->> pdp: Download Snapshot
+    pdp ->> factdb: Load Snapshot
+    cloud ->> factdb: Send Updates since Snapshot
+    pdp ->> factdb: Query Data
+    pdp ->> opa: Evaluate Policy using the Data
+```
+
+
+## Troubleshooting & Debugging
```
gemanor, line 153: Please add some intro here that state something like: Running the PDP local facts give you more responsibility in Ops so here are some advice to help you debug the PDP and data behavior

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+### Refreshing Data
+Every time the PDP boots, it downloads the latest snapshot from the Permit Cloud and stores it in the `/app/backup/` directory.
+
+By restarting the PDP, you can refresh the FactDB data with the most up-to-date data from the Permit Cloud.
```
gemanor, line 158: Better to state us: If you want to refresh the PDP data, just restart it and get the most...

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+### Extracting Data
+PDP with FactDB stores data in the `/app/backup/` directory. You can inspect the data by mounting the volume to your local machine.
```
gemanor, line 161: Should be more like:
To extract the FactsDB store in the PDP, just mount the `app/backup` volume in your local machine and use the data locally.

gemanor (changes requested):
Great start, thanks for the initiative 🚀

Left some comments, mostly about structuring/styling.

I still think that you need to put some metrics in the TL'DR to make people want to use it.

docs/concepts/pdp/factdb.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+With this setting enabled, you can run both FactDB PDPs with `PDP_FACTDB_ENABLED=true` and Legacy PDPs in the same environment.
+
+:::info
+We recommend you to migrate all your PDPs in the environment to FactDB.
+Please contact us on [Slack](https://io.permit.io/slack) for more information.
+:::
+
+
+### PDP Offline Mode with FactDB
```
obsd, line 96: Is it unique to FactsDB?
If not let's move it to offline mode section and add that factsDB is supporting this feature as well (if you want)

docs/concepts/pdp/factdb.mdx
danyi1212: Well, it is inside of the PDP subject, do we really need to link to its parent?

docs/concepts/pdp/factdb.mdx
danyi1212: Shortened the header

docs/concepts/pdp/factdb.mdx
danyi1212: Moved the warning to the enable guide

docs/concepts/pdp/factdb.mdx
danyi1212: Guess the change above gives it better context. Right after stating it is configured on a Permit Environment, it notifies it has to be on all PDPs in the environment

docs/concepts/pdp/factdb.mdx
danyi1212: I'll just remove it from this part, as we have a dedicated part for "both FactDB and Legacy". When you enable, it is enabled for all.

docs/concepts/pdp/factdb.mdx
danyi1212: Currently it overrides the settings, it does not merge it with the existing ones.

docs/concepts/pdp/factdb.mdx
danyi1212: The title right below that explains how to do it. I'll add a link to it anyway

docs/concepts/pdp/factdb.mdx
danyi1212: We recommend enabling it on all PDPs, that's why it's first. For advanced setup that want to migrate to it incrementally, you can also enable legacy as well and run PDPs with it both enabled and disabled.

I've attached an info tag as well to address that we recommend enabling it on all PDPs

docs/concepts/pdp/factdb.mdx
danyi1212: WDYM? When a PDP boots, it communicates with all other PDPs in the environment. 

Changed "the PDP" to "a PDP"

docs/concepts/pdp/factdb.mdx
danyi1212: But that's not a convincing thing to say (and is not really accurate🙈) 
The troubleshooting is mainly to make readers familiar and feel safe tinkering and experiencing it for themselves, and for us to link to on support

docs/concepts/pdp/factdb.mdx
danyi1212: I've shortened it and hopefully made it clearer

docs/concepts/pdp/factdb.mdx
danyi1212: No, but @orweis commented that we should explain how it works with Offline Mode, and I agree that we should emphasis on it here as people are likely going to question that

gemanor (approved):

PR #452, opened 2024-11-27
Added PDP_DATA_UPDATE_CALLBACKS config

danyi1212:


docs/concepts/pdp/configuration.mdx
```diff
@@ -264,6 +264,31 @@ Default: `<green>{time}</green> | {process} | <blue>{name: <40}</blue>|<level>{l
...
+      "data": {
+        "key": "value"
+      }
+    }
+  }
+]
+```
+The default config is: `{"method": "POST"}`.
+
+For more information, see the [CallbackEntry](https://github.com/permitio/opal/blob/master/packages/opal-common/opal_common/schemas/data.py#L123) interface.
```
omer9564, line 291: missing danger that having slow API receiving the callbacks might affect the PDP synchronization time

omer9564 (changes requested):

docs/concepts/pdp/configuration.mdx
```diff
@@ -264,6 +264,31 @@ Default: `<green>{time}</green> | {process} | <blue>{name: <40}</blue>|<level>{l
...
+
+Default: `[]`
+
+List of data update callbacks to be called when data is updated. For example:
+```json
+[
+  {
+    "url": "http://example.com/callback",
+    "key": "example",  // optional
+    "config": {  // optional
```
omer9564, line 277: this isn't a valid json

omer9564 (approved):
approved with comments

docs/concepts/pdp/configuration.mdx
danyi1212: Right, it's docs it does not have to be valid 🙂

PR #456, opened 2024-12-02
New Custom CodeBlock Component

filipermit:
image


src/codeFiles.json
```diff
@@ -0,0 +1,32 @@
+[
```
gemanor, line 1: This can be removed now (even with static import, there's no reason to merge this file)

scripts/generateCodeFiles.js
```diff
@@ -0,0 +1,18 @@
+const fs = require("fs");
```
gemanor, line 1: No need for this file anymore

package.json
```diff
@@ -38,12 +39,14 @@
     "docusaurus-plugin-hotjar": "^0.0.2",
     "docusaurus-plugin-sass": "^0.2.2",
     "file-loader": "^6.2.0",
+    "glob": "^11.0.0",
```
gemanor, line 42: Redundant with the dynamic

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
+    "start": "node scripts/generateCodeFiles.js && docusaurus start",
```
gemanor, line 7: Can revert

package.json
```diff
@@ -15,7 +15,8 @@
     "lint": "eslint --ignore-path .gitignore .",
     "format": "prettier --ignore-path .gitignore --write \"**/*.{ts,tsx,css,scss,js,mdx}\"",
     "hyperlink": "hyperlink  ./build --check-anchors  --sources ./docs",
-    "redirect-lint": "node checkRedirects"
+    "redirect-lint": "node checkRedirects",
+    "update-sdk": "node scripts/generateCodeFiles.js"
```
gemanor, line 19: Redundant

src/components/code-block/CodeBlock.js
```diff
@@ -0,0 +1,128 @@
+import React, { useEffect, useState } from "react";
+import PropTypes from "prop-types";
+import { Prism as SyntaxHighlighter } from "react-syntax-highlighter";
+import { dracula } from "react-syntax-highlighter/dist/esm/styles/prism";
+import "./CodeBlock.css";
+
+const context = require.context("!raw-loader!../../sdks", true);
```
gemanor, line 7: I'm not sure, but maybe we can even move it to a global file so it will read only once (although since Docusaurus is compiled to be static, there's no real reason for that)

gemanor (changes requested):
Great job!
Let's revert the tabs and codeblocks to the Docusaurus native ones, and then we can merge it.

docs/how-to/enforce-permissions/bulk-check.mdx
```diff
@@ -14,155 +15,37 @@ For some use cases, you might need to perform multiple `check()` calls at once.
...
 For example, assuming we have an API endpoint that performs multiple actions on a resource, we can use the `bulkCheck` function to check all the permissions at once.
 
-```js
-const permitted = await permit.bulkCheck([
-    { user, "create", "document" },
-    { user, "edit", "profile"}
-]);
-```
-
+<CodeBlock folderPath="/bulk-check/latency-applications" />
```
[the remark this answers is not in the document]
filipermit: I don't know the SDK implementation for Go or Java, and I wouldn't want to guess, but it does work.

src/components/code-block/CodeBlock.js
filipermit: We could do it, but I don't think it gives much performance improvement at all, and it's only used for the code block - I don't think we will be generating a tree of files anywhere - so for this reason I think it's better to keep it in the file. 

If we ever extend the functionality of this, then I'd say we move it into the utils. It's little work.

PR #463, opened 2024-12-18
fix policy_guard doc

maya-barak:


docs/manage-your-account/policy-guard/policy_guard_api.mdx
```diff
@@ -79,12 +84,15 @@ curl -X DELETE 'https://api.permit.io/v2/policy_guards/scopes/{policy_guard_scop
 curl -X POST 'https://api.permit.io/v2/policy_guards/scopes/{policy_guard_scope_id}/rules' \
 -H 'authorization: Bearer API_SECRET_KEY' \
 --data-raw '{
-    "resource": "documents",
-    "role": "admin",
-    "action": "create",
+    "resource_key": "documents",
```
RazcoDev, line 87: Add an example for ABAC as well

RazcoDev (approved):

docs/manage-your-account/policy-guard/policy_guard_api.mdx
```diff
@@ -79,12 +84,15 @@ curl -X DELETE 'https://api.permit.io/v2/policy_guards/scopes/{policy_guard_scop
...
+    "role_key": "admin",
+    "action_key": "create",
     "is_allowed": true
 }'
 ```
+:::note Info
+We use `key` to refer to the resource, role, and action in the examples above. The `key` is a unique identifier for each object, enabling its reference across different projects within the Policy Guard scope.
```
RazcoDev, line 94: You should say that we create and guard objects by key, as IDs are unique per environment, not Keys like you wrote.

PR #468, opened 2024-12-22
Update getuserpermission section - add new fucntino getUserPermission…

gideonsmila:
…sFromOPA

### DIFF
--- a/docs/how-to/enforce-permissions/user-permissions.mdx
+++ b/docs/how-to/enforce-permissions/user-permissions.mdx
@@ -163,8 +163,13 @@ UserPermissions permissions = permit.getUserPermissions(
 
 ##  Get user permissions directly from opa
 
-:::info Info
+:::info Note
 This feature is not supported when using FactDB and should not be used with FactDB enabled on the PDP
+:::
+
+:::info Info
+When experiencing high load, it may be more efficient to call the OPA engine directly from the SDK.
+Please note that to enable this, you will need to expose the OPA port (8181).
 :::
 
 ```java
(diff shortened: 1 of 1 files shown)


docs/how-to/enforce-permissions/user-permissions.mdx
```diff
@@ -160,7 +160,40 @@ UserPermissions permissions = permit.getUserPermissions(
 );
 
 ```
-</TabItem>
 
+##  Get user permissions directly from opa
+
+:::info Info
+Not Support FactDB
```
omer9564, line 167:
```suggestion
This feature is not supported when using FactDB and should not be used with FactDB enabled on the PDP
```

docs/how-to/enforce-permissions/user-permissions.mdx
```diff
@@ -160,7 +160,40 @@ UserPermissions permissions = permit.getUserPermissions(
...
+Context context = new Context();
+context.put("enable_abac_user_permissions", new Boolean(true));
+
+UserPermissions permissions = permit.getUserPermissionsFromOPA(
+	new GetUserPermissionsQuery(
+		User.fromString("john@doe.com"), // user key
+		null, // tenants filter is not required for ABAC
+		Arrays.asList("document", "__tenant"), // resource types is always required for ABAC, __tenants is required to not ignore RBAC-based permissions
+		null, // resources not required
+		context,
```
omer9564, line 191: this makes your example harder to understand ( I guess because you copied it from the ABAC example ). 
copy the most simple example here to demonstrate only the relevant function

docs/how-to/enforce-permissions/user-permissions.mdx
```diff
@@ -160,7 +160,40 @@ UserPermissions permissions = permit.getUserPermissions(
 );
 
 ```
-</TabItem>
 
+##  Get user permissions directly from opa
+
```
omer9564, line 165: add content here, something like
When having high load, it might be more performant to call the OPA engine directly from the SDK.
Note that to do that you'll also have to expose OPA port

omer9564 (changes requested):

PR #473, opened 2024-12-31
Timeline integration

filipermit:
image


src/components/video/Video.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const VideoPlayer = ({ src }) => {
+  return (
+    <video controls autoPlay loop className="w-full h-auto block">
```
gemanor, line 5: If we want people to use this component, we need to make it configurable (probably with default values), not all the videos should be in loop/autoplay. Please make it configurable.

gemanor (changes requested):
Great start. Please address the comments and merge it.

PR #490, opened 2025-01-21
Resource user condition

gemanor:


docs/how-to/build-policies/abac/building-abac-policy.mdx
```diff
@@ -111,6 +111,11 @@ Our conditions for this set will be setting **time** using the comparison **grea
 
 ![Resource Set Example](/img/abac/resource-set-1.png)
 
+### Compare Resource and User Attributes
```
filipermit, line 114: Just a little wording change:

Sometimes, you may want to create conditions that compare resources with user attributes, such as matching a resource's risk score to a user's risk score. In Permit, this can be achieved using the `(ref)` operators. Here's an example of such a condition:

docs/how-to/build-policies/abac/building-abac-policy.mdx
```diff
@@ -111,6 +111,11 @@ Our conditions for this set will be setting **time** using the comparison **grea
...
+In some cases, you'd like to create conditions that compare resource and user attributes, such as comparing a resource and user risk score. This comparison is available in Permit using the `(ref)` operators. Here's an example of such a condition:
```
gemanor, line 115:
```suggestion
Sometimes, you may want to create conditions that compare resources with user attributes, such as matching a resource's risk score to a user's risk score. In Permit, this can be achieved using the `(ref)` operators. Here's an example of such a condition:
```

docs/how-to/build-policies/abac/building-abac-policy.mdx
gemanor: Done

PR #518, opened 2025-02-23
Add Support for Private Browsing Mode

shuvyA:
image

### DIFF
--- a/docs/embeddable-uis/element-login.mdx
+++ b/docs/embeddable-uis/element-login.mdx
@@ -632,6 +632,10 @@ permit.elements.logout();
 ---
 ### Support for private browsing mode (EAP)
 
+:::info IMPORTANT
+This feature is available from version permitio/permit-js@0.5.2
+:::
+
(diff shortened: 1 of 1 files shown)


docs/embeddable-uis/element-login.mdx
```diff
@@ -629,6 +629,64 @@ authentication solution.
 permit.elements.logout();
 ```
 
+---
+### Support for Private Browsing Mode
+
+Permit Element is now compatible with private browsers, such as Chrome Incognito Mode and Safari Private Browsing.
```
obsd, line 635: rm Private Browsing

docs/embeddable-uis/element-login.mdx
```diff
@@ -629,6 +629,64 @@ authentication solution.
...
+
```
obsd, line 636: To do that you need to change a few things from the regular cookie-based permit elements

obsd (approved):

PR #530, opened 2025-03-09
chore: migrate to ESM and fix all breaking changes to prepare for uprgade to `Docusaurus v3.7`

asr2003:


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "my-website",
   "version": "0.0.0",
```
gemanor, line 3:
```suggestion
  "version": "0.0.1",
```

asr2003:
@ologbonowiwi Any suggestions on this build failure error. 

I debugged and issue is that docusaurus pulling some plugins of JS and it making type issues of `resolveNotWeak` etc.. which aren't compatible of ESM and  I tried by patching the package locally which isn't supported by ESM but it still pulling more errors to patch. And if we go non ESM the issue is that mdx react fully removed their support of non-ESM in their latest versions and docusaurus need mdx react latest version as dependency. What I am thinking we go fully support the ESM of our codebase better approach I guess

Any other things i am missing, please let me know and i will make sure complete it

asr2003:
Update: The ESM issue has been fixed but a minor issue left of is it still referencing to the cache builds even though i tried this to clear everything and build from scratch
`
 rm -rf .docusaurus/ build/ node_modules/.cache && npx docusaurus clear && npx docusaurus build
`

PR #535, opened 2025-03-26
Add PDP_LOCAL_FACTS_TIMEOUT_POLICY and update local-facts-uploader documentation

danyi1212:


sidebars.js
```diff
@@ -248,7 +248,11 @@ const sidebars = {
             type: "doc",
             id: "concepts/pdp/overview",
           },
-          items: ["concepts/pdp/configuration", "concepts/pdp/factdb"],
+          items: [
+            "concepts/pdp/configuration",
+            "concepts/pdp/factdb",
```
omer9564, line 253: I think we need to delete this docs

danyi1212: Opened a ticket for it
https://linear.app/permit/issue/PER-12068/remove-docs-for-factdb

docs/concepts/pdp/configuration.mdx
```diff
@@ -270,6 +270,14 @@ Default: `<green>{time}</green> | {process} | <blue>{name: <40}</blue>|<level>{l
 
 Change this parameter to change the log format (e.g. remove colors).
 
+#### PDP_LOCAL_FACTS_TIMEOUT_POLICY
```
orweis, line 273: This also requires a PDP min version no ?

docs/how-to/manage-data/local-facts-uploader.mdx
```diff
@@ -1,282 +1,352 @@
 ---
 sidebar_position: 3
-title: Send Critical Updates
+title: Proxy Facts via PDP
```
orweis, line 3: I really think we should stick with critical updates as a name for this.
Focusing on what it means for the customer, and less of what it technically does...

danyi1212: But it's not the name of the feature in the code, and "Critical Updates" meaning is vague and increases the confusion further

docs/concepts/pdp/configuration.mdx
orweis: Also we should specify which is the default

orweis (commented):
Great and consistent :)

orweis (approved):
Great and consistent :)

PR #544, opened 2025-04-09
refactor: improve playwright usage/pipeline feedback and remove unnecessary `tailwind` package

wesleymatosdev:
- **refactor(tests): use js playwright and fix eslint errors**
- **refactor: improve logging to show test progress**


argo_utils.js
ologbonowiwi, line 1: having this file not clear that was test-related would be bad for long-term maintenance

playwright.config.js
ologbonowiwi, line 1: js over ts files to avoid the unnecessary complexity

tests/screenshot.spec.js
```diff
@@ -0,0 +1,78 @@
...
+// Calculate and display progress of the running tests
+function calculateProgress(index, total, pathname) {
+  const percentage = Math.floor((index / total) * 100);
+  const progressBarWidth = 30; // Width of the progress bar in characters
+  const filledSquares = Math.floor((percentage / 100) * progressBarWidth);
+  const emptySquares = progressBarWidth - filledSquares;
+  const progressBar = '█'.repeat(filledSquares) + '░'.repeat(emptySquares);
+
+  console.log(`Progress: [${progressBar}] ${percentage}% (${index + 1}/${total}) - Testing: ${pathname}`);
+}
```
ologbonowiwi, line 47: improved view so the CI now looks like this 
image

argo_utils.js
ologbonowiwi, line 1: having this file not clear that was test-related would be bad for long-term maintenance. also too short and straightforward to need its own file. will never be reused.

playwright.config.js
ologbonowiwi, line 1: using `.js` instead of `.ts` to avoid unnecessary complexity

argo_utils.js
ologbonowiwi, line 1: having this file not clear that was test-related would be bad for long-term maintenance. also too short and straightforward to need its own file. will never be reused.

playwright.config.js
ologbonowiwi, line 1: using `.js` instead of `.ts` to avoid unnecessary complexity

package.json
```diff
@@ -54,7 +58,6 @@
     "react-syntax-highlighter": "^15.6.1",
     "remark-gfm": "^3.0.1",
     "sass": "^1.81.0",
-    "tailwind": "^4.0.0",
```
ologbonowiwi, line 57: this is some shadowy package that was likely installed by mistake. [tailwindcss](npmjs.com/package/tailwindcss) is the right package (and it is still installed)

PR #561, opened 2025-05-29
on-prem docs skeleton

EliMoshkovich:


ocap-kirk:
Is there any risk in letting the world know exactly what all of our services look like? The level of detail is great for what a customer who is actually implementing is doing, but does the whole world need that level of detail?

The quick start guide seems like it's sufficent enough without the full installation details like you put in the Installation guide. @orweis what's your thoughts? Any risk, because it shows all services and names and default ports.

orweis:
I don't believe in security by obscurity.
The more light the better.
I don't see an issue here

ocap-kirk:
Looks good. Nice work on this @EliMoshkovich

ocap-kirk (approved):

PR #599, opened 2025-12-03
Add Envoy Proxy integration documentation for Permit.io

omer9564:
Adds a guide for integrating Envoy's External Authorization (`ext_authz`) filter with a Permit PDP, using a GitOps-managed custom Rego entrypoint, and registers it in the Proxies & API Gateways sidebar alongside the AWS API Gateway, Kong and NGINX guides.

## What the guide covers

1. **Prerequisites** — Envoy with the `ext_authz` filter, a running PDP on 0.9.10 or higher, and GitOps configured for the environment.
2. **GitOps setup** — pointing at the CLI flow (`/how-to/permit-cli/permit-cli-gitops`) and the manual flow (`/integrations/gitops/github`).
3. **Envoy entrypoint policy** — a `custom/envoy_entrypoint.rego` example that reads the request from `input.attributes.request.http`, builds a check input, and delegates to `data.permit.root.allow`.
4. **PDP OPA plugin config** — enabling `envoy_ext_authz_grpc` via `PDP_OPA_PLUGINS`, with `addr` (gRPC listen address) and `path` (the OPA data path of the entrypoint), alongside the required `permit_graph` plugin.
5. **Envoy config** — an `ext_authz` HTTP filter plus the matching cluster definition pointing at the PDP's gRPC port.
6. **URL Mapping** — how URL Mapping relates to the Envoy flow, with links to the simple and regex URL-mapping guides.
7. **Audit Log ingestion** — returning an object result with a `permitio_metadata` field so Envoy decisions populate the Audit Log's user, action, resource, tenant and decision columns. Includes the full field-to-column table, the precedence rule (input fields win; `permitio_metadata` is a fallback), the requirement that the OPA path stay under the `permit/` prefix, and a note that these rows only show up under the **All** query type rather than the Check filter.
8. **Example repository** — placeholder for a future end-to-end sample.

## Files changed

- `docs/integrations/gateways/envoy.mdx` — new guide (257 lines).
- `sidebars.js` — adds `integrations/gateways/envoy` to the Proxies & API Gateways category.

## Notes

Documentation only — no code, config or build changes. Nothing outside the new page and its sidebar entry is touched.


zeevmoney:
Re-open when relevant.

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+# This block will show how to resolve the incoming URL and method
+# using Permit URL Mapping and then perform a `permit.check`.
+```
+
+1. **Define URL Mappings** in the Permit UI:
+   - For simple templated URLs, follow: [Simple URL Mapping Check](/how-to/enforce-permissions/url-mapping/url-mapping-check)
+   - For advanced patterns, follow: [Regex URL Mapping Check](/how-to/enforce-permissions/url-mapping/regex-url-mapping-check)
+
+2. **Regex mappings and captured group**:
+   - When using **regex URL mapping**, the **first captured group** in the regex will be used as the **resource key** in the `permit.check` evaluation.
```
zeevmoney: Regex URL mapping does not use the first captured group as the resource key

This contradicts both the PDP implementation and the sibling doc.

In `PDP/horizon/enforcer/api.py`, the regex branch of `is_allowed_url` uses **named** capture groups, and they become resource **attributes** — not the resource key:

```python
if matched_mapping_rule.url_type == "regex":
    # For regex patterns, use only named capture groups
    pattern = re.compile(matched_mapping_rule.url)
    match = pattern.match(query.url)
    path_attributes = match.groupdict() if match else {}
...
allowed_query = AuthorizationQuery(
    user=query.user,
    action=matched_mapping_rule.action,
    resource=Resource(
        type=matched_mapping_rule.resource,   # <- resource comes from the mapping rule
        tenant=query.tenant,
        attributes=attributes,                # <- captured groups land here
    ),
    ...
)
```

`match.groupdict()` returns only *named* groups, so an unnamed first group is discarded entirely. The resource type is always `matched_mapping_rule.resource` — the resource configured on the mapping rule in the UI.

`docs/how-to/enforce-permissions/url-mapping/regex-url-mapping-check.mdx:36` already states this correctly: "Regex URL mapping uses named capture groups (e.g., `(?P<param_name>pattern)`) to extract values from the URL. These captured values are automatically passed as attributes to your policy."

**Suggestion:** Replace this bullet with the accurate behavior, matching the sibling doc:

> - **Regex mappings and captured groups**: when using regex URL mapping, **named** capture groups (`(?P<name>pattern)`) are extracted and passed to the policy as resource **attributes**. The resource itself is taken from the resource configured on the mapping rule, not from the URL.

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+Adjust the cluster name, address, and port to match your environment.
+Consult the [Envoy ext_authz docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter)
+for the full configuration options.
+
+---
+
+## 5. Using URL Mapping with Envoy
+
+You can leverage Permit **URL Mapping** so that Envoy only passes the raw URL and method, while the PDP resolves them into Permit resources and actions.
```
zeevmoney: URL Mapping is not reachable from the ext_authz path this guide sets up

This section promises that "the PDP resolves them into Permit resources and actions", but URL Mapping resolution lives entirely in the PDP's Python API layer, which the Envoy gRPC path never touches.

The whole matching pipeline — rule selection, `{var}` extraction, regex named-group extraction, query-param handling — is Python, in `PDP/horizon/enforcer/api.py::is_allowed_url` and `PDP/horizon/enforcer/utils/mapping_rules_utils.py`. It is only reachable via `POST /allowed_url` on the PDP HTTP API (port 7766).

The integration this guide configures does not go through that layer at all. `envoy_ext_authz_grpc` is an OPA plugin registered in the custom OPA binary (`permit-opa/main.go:300`) and served from OPA's own listener on `:9191`. Envoy talks straight to OPA and evaluates `data.permit.envoy`; the PDP's FastAPI app is never in the request path.

OPA only holds the raw rules at `data.mapping_rules.all` (that is what `is_allowed_url` fetches before doing the work in Python). There is no Rego helper that matches a URL and method against them, which is why both code blocks in this section are placeholders — there is nothing to fill them with today.

**Suggestion:** Either drop this section until Rego-side URL-Mapping resolution exists, or rewrite it to say plainly what a reader can do now: the Envoy entrypoint policy has to map the request to a resource and action itself in Rego (as sections 2 and 6 already show), and URL Mapping applies to the PDP's `/allowed_url` HTTP API rather than to the ext_authz gRPC path.

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+     user_key := http_request.headers["x-user-id"]
+     tenant_key := http_request.headers["x-tenant-id"]
+
+     check_input := {
+       "user": {
+         "key": user_key,
+       },
+       "action": method,
+       "resource": {
+         "type": url,
```
zeevmoney: Using the raw URL path as `resource.type` means this example can never return allow

`url` here is `input.attributes.request.http.path`, so it is a full request path such as `/documents/123`. Permit resource keys cannot contain `/`:

```python
# permit-backend/v2/backend/permit_backend/schemas/fields.py:42
KEY_REGEX = r"^[A-Za-z0-9\-_]+$"
```

No resource can ever have a key that matches a URL path, so `data.permit.root.allow` evaluates to `false` on every request and the entrypoint always denies. This is the only complete, runnable example in the guide (section 5 is a placeholder), so a reader who copies it gets a gateway that blocks all traffic with no indication of why.

The same construct is repeated in the section 6 example on line 208.

The Kong guide already documents the correct shape — `kong.mdx:91-94`: "By default, `/` will be mapped to the 'index' resource, and other routes to the first element in their paths. `/repo` will be mapped to the `repo` resource."

Note `"action": method` is fine — `GET`/`POST` satisfy `KEY_REGEX`, and the Terraform sample in `integrations/gateways/overview.mdx` uses HTTP methods as action keys.

**Suggestion:** Derive a resource key from the path instead of using it verbatim, and say in the surrounding prose that the resulting key must match a resource configured in Permit.

**Example:**

```rego
# take the first path segment as the resource key: /documents/123 -> "documents"
segments := split(trim_prefix(http_request.path, "/"), "/")
resource_key := segments[0]

check_input := {
  "user": {"key": user_key},
  "action": method,
  "resource": {
    "type": resource_key,
    "tenant": tenant_key,
  },
}
```

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+1. Send the request details (URL, method, user identity, tenant, etc.) to the PDP.
+2. Use URL Mapping (if enabled) to resolve the request into a resource/action.
+3. Evaluate your Permit policies via `permit.check` and return **allow/deny** to Envoy.
+
+---
+
+## 6. Ingest Envoy decisions into Permit Audit Logs
+
+Envoy's ext_authz input is **not** a `permit.check` payload. The PDP still records a decision log for the Envoy query (`permit/envoy`), but the Audit Log page cannot fill **user**, **action**, **resource**, **tenant**, or **decision** from that Envoy-shaped input.
```
zeevmoney: The Decision column *is* filled for the boolean entrypoint

The claim that the Audit Log cannot fill **decision** is not accurate. User, action, resource and tenant are read from the input and genuinely cannot be resolved from an Envoy-shaped input — but the decision is read from the *result*, not the input:

```python
# permit-backend/v2/common/permit_common/engines/opa.py:187-199
def get_decision(self) -> bool | None:
    if self.result is None:
        return None
    if (result_type := type(self.result)) is dict:
        decision = self._get_decision_by_query()
        if decision is not None:
            return decision
        # Lowest precedence fallback: result.permitio_metadata.allow
        return jsonpath_first("$.permitio_metadata.allow", self.result, bool)
    elif result_type is bool:
        return self.result
    return None
```

The section 2 entrypoint (`default envoy = false`) produces a raw boolean result — `opa-envoy-plugin` logs the decision verbatim (`opa/decisionlog/decision_log.go`: `x = result.Decision; info.Results = &x`) — so `result_type is bool` fires and the Decision column is populated without any `permitio_metadata`.

This also affects whether the row appears at all: `AuditLogModel.is_empty()` (`v2/backend/permit_backend/schemas/schema_audit_logs/interface.py:153-162`) passes as soon as `decision is not None`, so boolean-entrypoint rows already reach the Audit Log — just with empty user/action/resource/tenant columns.

**Suggestion:** Drop **decision** from the list and describe the boolean case accurately:

> Envoy's ext_authz input is **not** a `permit.check` payload. The PDP records a decision log for the Envoy query (`permit/envoy`) and fills the **Decision** column from the boolean result, but the Audit Log page cannot fill **user**, **action**, **resource** or **tenant** from that Envoy-shaped input.

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+Return an object from the Envoy entrypoint (the OPA Envoy plugin uses the `allowed` field for allow/deny):
+
+```rego
+package permit
+
+import data.permit.root as check
+import input.attributes.request.http as http_request
+
+default envoy := {"allowed": false}
```
zeevmoney: The default value has no `permitio_metadata`, so default-denied requests vanish from the Audit Log

When the rule body fails — a missing `x-user-id` or `x-tenant-id` header, an unparseable path — this default fires and the result is `{"allowed": false}` with no `permitio_metadata`. Every Audit Log field then resolves to `None`:

- `get_decision()` — result is a dict, so `_get_decision_by_query()` runs first and does `self.result.get("allow")` for the `permit/envoy` query. The key is `allowed`, not `allow`, so that returns `None`, and the `$.permitio_metadata.allow` fallback returns `None` too.
- `get_user_identifier()` / `get_action()` / `get_resource_type()` / `get_tenant()` — nothing in the input, nothing in `permitio_metadata`.

`AuditLogModel.is_empty()` then returns `True`, and the enricher drops the row (`v2/decision_logs_enricher/permit_dl_enricher/main.py:180-190`, `if audit_log.is_empty(): continue`).

The result is that exactly the denials an operator most wants to see — requests rejected because identity headers were missing — are the ones silently absent from the Audit Log. This is worth calling out in the guide, because the section 2 boolean entrypoint does not have this problem (a bare `false` result still yields a Decision).

**Suggestion:** Put the metadata on the default value too, so the default-deny path still produces an ingestable row.

**Example:**

```rego
default envoy := {
  "allowed": false,
  "permitio_metadata": {"allow": false},
}
```

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+    lb_policy: round_robin
+    load_assignment:
+      cluster_name: permit_pdp_ext_authz
+      endpoints:
+        - lb_endpoints:
+            - endpoint:
+                address:
+                  socket_address:
+                    address: pdp
+                    port_value: 9191
```
zeevmoney, line 165: Guide never says to publish port 9191, which the PDP does not expose by default

Envoy is pointed at `pdp:9191`, but nothing earlier in the guide makes that port reachable. The PDP image only declares:

```dockerfile
# PDP/Dockerfile:222-225
# 7000 pdp port
# 7001 horizon port
# 8181 opa port
EXPOSE 7000 7001 8181
```

Setting `PDP_OPA_PLUGINS` starts the gRPC listener inside the container, but that is a separate step from making it reachable from Envoy. The PDP Helm chart documents this explicitly as a second required change:

```yaml
# PDP/charts/pdp/values.yaml:21-25
  # Example - expose Envoy gRPC ext_authz port (requires PDP_OPA_PLUGINS env var above):
  #   additionalPorts:
  #     - name: grpc
  #       port: 9191
  #       targetPort: 9191
```

Without it the Kubernetes Service has no 9191 port and every ext_authz call fails to connect — with `failure_mode_allow` at its default of `false`, that means Envoy denies all traffic.

**Suggestion:** Add the port-publishing step to section 3, next to `PDP_OPA_PLUGINS`, covering both deployment styles:

> The gRPC listener also has to be reachable from Envoy:
>
> - **Docker:** add `-p 9191:9191` to the `docker run` command (or a `ports` entry in Compose).
> - **Helm:** add the port to the PDP Service via `pdp.additionalPorts`:
>
> ```yaml
> pdp:
>   additionalPorts:
>     - name: grpc
>       port: 9191
>       targetPort: 9191
> ```

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+   custom/envoy_entrypoint.rego
+   ```
+
+2. Define an entrypoint rule that Envoy's ext_authz integration will call.
+   The PDP OPA plugin will evaluate the data document at the path you configure (see the `PDP_OPA_PLUGINS` section below).
+
+   A simplified example of an Envoy entrypoint policy might look like this:
+
+   ```rego
+   package permit
```
zeevmoney: Bare `package permit` diverges from the convention documented for the `custom/` folder

Every other file Permit generates or ships uses a subpackage: `permit.root`, `permit.policies`, `permit.rbac`, `permit.abac`, `permit.rebac`, and the seeded `custom/root.rego` uses `permit.custom`. `docs/integrations/gitops/custom_policy.mdx:42` states this as the convention for custom code.

Declaring bare `package permit` from a file under `custom/` does work today — OPA derives the package from the declaration rather than the directory, no other file claims `data.permit` directly, and the policy synchronizer does not validate package names on custom files. But it makes this the only file in the repo that writes into the `data.permit` root namespace, which is where Permit's generated policies live, so any future generated rule named `envoy` would silently collide.

**Suggestion:** Keep it in the custom namespace and adjust the plugin path to match.

**Example:**

```rego
package permit.custom.envoy
```

with:

```bash
export PDP_OPA_PLUGINS='{"permit_graph":{},"envoy_ext_authz_grpc":{"addr":":9191","path":"permit/custom/envoy"}}'
```

This still satisfies the `permit/` prefix requirement for Audit Log ingestion noted in section 6.

zeevmoney (changes requested):
Changes requested — 3 HIGH, 3 MEDIUM, 1 LOW.

Every factual claim in the guide was checked against the PDP, permit-opa, opa-envoy-plugin and permit-backend sources. Most hold up: the `PDP_OPA_PLUGINS` / `permit_graph` requirement, the 0.9.10 minimum, the `envoy_ext_authz_grpc` plugin name, the `path` to `data.permit.*` mapping, the `permit/` prefix ingestion filter, all seven `permitio_metadata` field-to-column mappings, the input-wins precedence rule, and the Rego v0 syntax. The findings below are the claims that do not.

Blocking:
- **HIGH** `docs/integrations/gateways/envoy.mdx:171` — regex URL mapping does not use the first captured group as the resource key; named groups become resource attributes and the resource comes from the mapping rule, per `horizon/enforcer/api.py` and the existing `regex-url-mapping-check.mdx`
- **HIGH** `docs/integrations/gateways/envoy.mdx:158` — URL Mapping resolution is implemented only in the PDP's Python API (`/allowed_url`), which the ext_authz gRPC path never reaches; the section describes a flow that is not achievable today
- **HIGH** `docs/integrations/gateways/envoy.mdx:75` — `resource.type` is set to the raw URL path, which cannot match any Permit resource key under `KEY_REGEX`, so the guide's only complete example always denies

Non-blocking:
- **MEDIUM** `docs/integrations/gateways/envoy.mdx:187` — the Decision column is populated for the boolean entrypoint; `get_decision()` returns raw boolean results directly
- **MEDIUM** `docs/integrations/gateways/envoy.mdx:199` — the default result carries no `permitio_metadata`, so `is_empty()` drops default-denied requests from the Audit Log entirely
- **MEDIUM** `docs/integrations/gateways/envoy.mdx:147` — port 9191 is not in the PDP image's `EXPOSE` and needs `pdp.additionalPorts` on Helm; the guide never says to publish it
- **LOW** `docs/integrations/gateways/envoy.mdx:52` — bare `package permit` diverges from the `permit.custom` convention documented for the `custom/` folder

Details are in the inline comments on each line.

Two open threads from an earlier review are still valid and unaddressed: the missing `http2_protocol_options` on the gRPC cluster (line 138) and the placeholder content (lines 164 and 255). Not duplicated here.

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+      grpc_service:
+        envoy_grpc:
+          cluster_name: permit_pdp_ext_authz
+        timeout: 0.5s
+
+clusters:
+  - name: permit_pdp_ext_authz
+    type: logical_dns
+    connect_timeout: 0.25s
+    lb_policy: round_robin
```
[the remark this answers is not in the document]
omer9564: Fixed in e508d961: added `http2_protocol_options: {}` on the gRPC cluster so ext_authz speaks HTTP/2. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+## 5. Using URL Mapping with Envoy
+
+You can leverage Permit **URL Mapping** so that Envoy only passes the raw URL and method, while the PDP resolves them into Permit resources and actions.
+
+```rego
+# Placeholder: URL Mapping Rego example for Envoy will be added here.
+# This block will show how to resolve the incoming URL and method
+# using Permit URL Mapping and then perform a `permit.check`.
+```
```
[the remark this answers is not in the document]
omer9564: Fixed in e508d961: removed the URL Mapping TODO blocks and the example-repository placeholder. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: dropped the incorrect first-capture-group claim along with the URL Mapping section. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: removed the URL Mapping section. The overview now states that URL Mapping is `POST /allowed_url` on the PDP HTTP API and is not used by ext_authz gRPC. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: resource type is now the first path segment (`/documents/123` → `documents`), which must match a Permit resource key. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: the guide now says the Decision column is filled from the boolean result; `permitio_metadata` is for user/action/resource/tenant. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: the default result now includes `permitio_metadata.allow: false` so default-denied requests are ingested. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: documented publishing port 9191 via Docker `-p 9191:9191` and Helm `pdp.additionalPorts`. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

docs/integrations/gateways/envoy.mdx
omer9564: Fixed in e508d961: switched to `package permit.custom.envoy` and `path: permit/custom/envoy`. Thanks!

https://github.com/permitio/docs/commit/e508d961ff8462e906b5bdea6adec24b9ebe71bb

omer9564:
@zeevmoney Addressed the requested changes in e508d961 — dropped the unreachable URL Mapping section, fixed resource-key / package / audit-log claims, and documented HTTP/2 plus publishing port 9191. Replies are on the inline threads.

PR #623, opened 2026-04-19
Document gateway JWT for upstream MCP server authentication

dshoen619:
## Summary

Documents the gateway JWT feature added in [agent-security#162](https://github.com/permitio/agent-security/pull/162) and the Redis-backed shared-key follow-up in [agent-security#172](https://github.com/permitio/agent-security/pull/172).

Adds a new **Upstream Authentication (Gateway JWT)** section to the MCP Gateway architecture page so users understand how the gateway proves request origin to upstream MCP servers, and gives them everything they need to verify on their side.

## What's documented

- **Purpose** — how the signed `X-Gateway-Auth` header lets upstream MCP servers reject traffic that didn't flow through the gateway (preventing bypass of authN/authZ/consent)
- **How it works** — a sequence diagram showing Ed25519 signing, lazy JWT refresh (re-sign within 30s of expiry), and the optional upstream JWKS verification path
- **JWT claims table** — `iss`, `sub`, `aud`, `exp`, `iat`, `nbf`, `jti`, `tenant`
- **Two headers coexist** — short table clarifying the roles of `Authorization: Bearer` (upstream OAuth) vs `X-Gateway-Auth: Bearer` (gateway identity proof)
- **Verification snippet** — ~10-line `jose` example for upstream servers, using `createRemoteJWKSet` so JWKS caching and key rotation are handled automatically
- **Key rotation note** — signing key is **cached in Redis** (encrypted at rest) and shared across all gateway replicas, so JWTs survive restarts and the JWKS endpoint is consistent across pods; keys rotate on roughly a 90-day cadence
- **Configuration** — `GATEWAY_JWT_TTL_SECONDS` env var (default 300) and pointer to the platform **Settings > Upstream Authentication (JWT)** panel

Also updated:

- **System Architecture Overview diagram** — Gateway → Upstream edge now labeled `Proxy MCP + X-Gateway-Auth JWT`, plus a dashed optional verification arrow back to the gateway's JWKS endpoint
- **Data Flow table** — Gateway → Upstream MCP row mentions the signed `X-Gateway-Auth` JWT with an anchor link to the new section

## Scope decisions

- Kept to the architecture page only (not platform or host-setup), since the feature is always-on with no required customer configuration on the gateway — upstream-side verification is the only user-visible surface
- Intentionally omitted Redis `SET NX` coordination, AES-256-GCM / KMS internals, and replica-count details — implementation specifics with no troubleshooting value for end users

## Test plan

- [ ] Verify the new section renders correctly (mermaid sequence diagram, tables, code block)
- [ ] Confirm the updated System Architecture Overview diagram renders with the new edges
- [ ] Check that the anchor link from the Data Flow table (`#upstream-authentication-gateway-jwt`) resolves correctly


docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+The gateway maintains an Ed25519 signing key that is cached in Redis and shared across all gateway replicas. The public key is exposed at `GET /.well-known/gateway-jwks.json` (unauthenticated, cached by clients for 5 minutes). On each upstream request, the gateway signs a short-lived JWT with the authenticated user's identity, the upstream URL as audience, and the tenant subdomain. The JWT is cached and reused until it is within 30 seconds of expiry, then re-signed — avoiding per-request signing overhead while ensuring long-lived MCP sessions always send a valid token.
+
+```mermaid
+sequenceDiagram
+    participant Client as MCP Client
+    participant GW as Gateway
+    participant JWKS as Gateway JWKS<br/>/.well-known/gateway-jwks.json
+    participant Up as Upstream MCP Server
+
+    Note over GW: Startup: generate ephemeral<br/>Ed25519 key pair
```
zeevmoney: Sequence-diagram startup note contradicts the shared-key model

`Note over GW: Startup: generate ephemeral Ed25519 key pair` describes the in-memory ephemeral path (`GatewayJwtSigner::new`), not the production path. Production calls `load_or_create`, which loads a shared key from Redis or atomically stores a new one with `SET NX` (`gateway/src/clients/gateway_jwt.rs`). The wording directly contradicts the prose two paragraphs above ("cached in Redis and shared across all gateway replicas"). The agent-security `gateway/README.md` diagram captions this correctly.

**Suggestion:** Match the README:

```
Note over GW: Startup: load shared Ed25519 key<br/>from Redis (or generate + store if none exists)
```

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+    GW-->>Client: MCP response
+```
+
+### JWT claims
+
+| Claim    | Value                       | Purpose                              |
+| -------- | --------------------------- | ------------------------------------ |
+| `iss`    | `agent-security-gateway`    | Identifies the issuer                |
+| `sub`    | Authenticated user ID       | Who the request is being made for    |
+| `aud`    | Upstream MCP server URL     | Intended recipient                   |
```
zeevmoney: `aud` row mis-states the claim value — canonicalization is silent

The signer runs `normalize_audience` (lowercases the host, strips all trailing slashes including the root `/`) before setting `aud`. The TS-side mirror `canonicalizeGatewayAudience` documents the rule explicitly. A customer whose configured upstream URL is `https://MCP.Example.com/` or `https://your-server.example.com/v1/` will pass the raw URL into their verifier's `audience` option and get a persistent verification failure with no obvious cause.

**Suggestion:** Replace the value cell and add the rule explicitly:

```
| `aud`    | Canonicalized upstream URL — host lowercased, all trailing slashes (including root `/`) stripped | Intended recipient                   |
```

Then add a short callout above the verifier snippet showing an example transform: `https://MCP.Example.com/` → `https://mcp.example.com`.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+| `exp`    | `now + TTL` (default 5 min) | Token expiry                         |
+| `iat`    | Current timestamp           | When the token was issued            |
+| `nbf`    | Current timestamp           | Not valid before this time           |
```
zeevmoney: `iat` and `nbf` are backdated 30 s, not "current timestamp" — verifier guidance missing

The signer sets `nbf = now - NBF_LEEWAY_SECS` (30 s) and `iat = nbf` (`gateway/src/clients/gateway_jwt.rs`, `sign_token`). The agent-security `gateway/README.md` table shows this correctly as `now - 30s (leeway)` and tells verifiers to allow ≥30 s of past skew. A verifier configured with strict clock tolerance can reject otherwise-valid tokens; the doc as written gives them no reason to widen tolerance.

**Suggestion:** Mirror the README and add a "Clock-skew leeway" note below the table.

```
| `iat`    | `now − 30 s` (clock-skew leeway) | When the token was issued     |
| `nbf`    | `now − 30 s` (clock-skew leeway) | Not valid before this time    |
```

Note: "Both `iat` and `nbf` are backdated 30 seconds to absorb clock skew between the gateway and upstream verifiers. Configure your JWT library with a clock tolerance of at least 30 seconds (e.g. `clockTolerance: 30` in `jose`)."

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+| `jti`    | Unique UUID                 | Unique ID (enables replay detection) |
```
zeevmoney: `jti` "enables replay detection" — `jose.jwtVerify` does not check `jti`

The Purpose cell sets the expectation that the verifier handles replay automatically. `jose.jwtVerify` (the library used in the snippet below) does not deduplicate `jti` values. Upstream developers reading this will think replay protection is built in.

**Suggestion:** Either drop the "(enables replay detection)" claim, or add a one-line note in the verifier section:

> Replay protection is opt-in: track recently-seen `jti` values for at least the JWT TTL plus your clock-skew window (e.g. a Redis SET with TTL = `GATEWAY_JWT_TTL_SECONDS + 30`).

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+
+```js
+import { createRemoteJWKSet, jwtVerify } from 'jose';
+
+const JWKS = createRemoteJWKSet(
+  new URL('https://<tenant>.agent.security/.well-known/gateway-jwks.json')
+);
+
+const { payload } = await jwtVerify(
+  req.headers['x-gateway-auth'].slice(7),  // strip 'Bearer '
```
zeevmoney: Verifier snippet crashes on missing header, doesn't validate the `Bearer` prefix, and has no clock tolerance

`req.headers['x-gateway-auth']` is `string | string[] | undefined`. If the header is absent — exactly the case the next paragraph says to reject — `.slice(7)` throws `TypeError: Cannot read properties of undefined`, producing a 500 instead of a 401. A doubled header returns an array (`["Bearer X", "Bearer Y"].slice(7) === []`), and a malformed value (e.g. `bearer X` lowercase, no prefix at all) is silently truncated and passed to `jwtVerify`. The snippet also defaults `clockTolerance` to 0 s, which is incompatible with the gateway's 30 s `nbf`/`iat` backdating when the verifier clock runs ahead.

**Suggestion:** Replace the call with a guarded version. Adds three lines but addresses all four issues:

```js
const auth = req.headers['x-gateway-auth'];
if (typeof auth !== 'string' || !auth.startsWith('Bearer ')) {
  return res.status(401).end();
}
const { payload } = await jwtVerify(auth.slice(7).trim(), JWKS, {
  issuer: 'agent-security-gateway',
  audience: 'https://your-mcp-server.example.com', // canonical: lowercase host, no trailing slash
  algorithms: ['EdDSA'],
  clockTolerance: 30,
});
```

(With the guard, the "~5 lines of code" framing on the line above is no longer accurate — consider "under 10 lines" or dropping the count.)

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+  JWKS,
+  {
+    issuer: 'agent-security-gateway',
+    audience: 'https://your-mcp-server.example.com',
+    algorithms: ['EdDSA'],
+  }
+);
+```
+
+Reject any request without a valid `X-Gateway-Auth` header. The `aud` claim must match the exact URL the gateway uses to reach your upstream server.
```
zeevmoney: "The `aud` claim must match the exact URL" — it's the canonicalized URL, not the exact one

The signer canonicalizes before signing (`normalize_audience`: lowercase host, strip all trailing slashes including the root `/`). A reader following this sentence literally — "match the exact URL the gateway uses to reach your upstream server" — will configure `audience` with the raw URL stored on the gateway side and get persistent 401s when that URL has a trailing slash or uppercase characters.

**Suggestion:**

> Reject any request without a valid `X-Gateway-Auth` header. The `aud` claim is the **canonicalized form** of your upstream URL — host lowercased, all trailing slashes (including the root `/`) stripped. Configure `audience` with the canonicalized form (e.g. `https://MCP.Example.com/` is canonicalized to `https://mcp.example.com`).

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+
+:::info Key rotation
+The signing key is cached in Redis (encrypted at rest) and shared across all gateway replicas, so JWTs remain valid across pod restarts and the JWKS endpoint returns the same key regardless of which replica serves it. Keys are expected to rotate on roughly a 90-day cadence. The `kid` header on each JWT identifies which key signed it — upstream servers should rely on their JWT library's JWKS caching, which automatically re-fetches the JWKS on unknown `kid`, making rotation transparent.
```
zeevmoney: Encryption-at-rest is conditional on vault; 90-day rotation is manual, not automatic

Two implementation drifts in this callout:

1. "encrypted at rest" — only when the token vault is enabled (`VAULT_ENABLED=true` + `AWS_KMS_KEY_ID`). Without vault, the key is stored as plaintext JSON in Redis (the gateway logs a warning). The agent-security `gateway/README.md` qualifies this correctly.
2. "Keys are expected to rotate on roughly a 90-day cadence" — the implementation only emits a `warn!` log + Prometheus metric after 90 days. Rotation is entirely manual: `redis-cli DEL gateway:jwt:signing_key` followed by a rolling pod restart. The wording reads as if rotation happens automatically.

**Suggestion:**

> The signing key is stored in Redis and shared across all gateway replicas, so JWTs remain valid across pod restarts and the JWKS endpoint returns the same key from any replica. When the token vault is enabled (`VAULT_ENABLED=true` + `AWS_KMS_KEY_ID`), the key is encrypted at rest with AES-256-GCM + AWS KMS envelope encryption; without vault it is stored as plaintext JSON. The gateway logs a warning after 90 days and emits a `gateway_jwt_signing_key_age_seconds` metric — rotation is manual: `redis-cli DEL gateway:jwt:signing_key` followed by a rolling pod restart. The `kid` header on each JWT identifies which key signed it; standard JWKS-aware libraries refetch automatically on an unknown `kid`, so rotation is transparent to verifiers.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,111 @@ The derived role then determines which tools are allowed:
...
+:::
+
+### Configuration
+
+There is nothing to configure on the gateway to enable this feature. The only optional tunable is the JWT lifetime:
+
+| Env var                   | Default | Description             |
+| ------------------------- | ------- | ----------------------- |
+| `GATEWAY_JWT_TTL_SECONDS` | `300`   | JWT lifetime in seconds |
```
zeevmoney: Configuration table is missing `GATEWAY_JWT_REQUIRE_VAULT` and TTL bounds

Two omissions vs. the implementation in `gateway/src/config.rs` (`GatewayJwtConfig`):

1. `GATEWAY_JWT_REQUIRE_VAULT` (default `false`) is the operator-facing toggle that enforces encryption-at-rest. Production deployments need to know this exists — it's the only way to prevent the plaintext-key fallback path called out in the Key rotation block.
2. `GATEWAY_JWT_TTL_SECONDS` is bounded: `60 ≤ ttl ≤ 3600`. Values outside that range fail config validation at startup. Operators tuning this need to know.

**Suggestion:**

```
| Env var                     | Default | Description                                                                                                       |
| --------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------- |
| `GATEWAY_JWT_TTL_SECONDS`   | `300`   | JWT lifetime in seconds (must be between 60 and 3600).                                                             |
| `GATEWAY_JWT_REQUIRE_VAULT` | `false` | Refuse to start if the token vault is not configured. Prevents plaintext private-key storage in Redis. Recommended `true` in production. |
```

zeevmoney (changes requested):
see comments.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — sequence-diagram note now matches the README: `Note over GW: Startup: load shared Ed25519 key<br/>from Redis (or generate + store if none exists)`. Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — `aud` row now reads "Canonicalized upstream URL — host lowercased, all trailing slashes (including root `/`) stripped". Added an explicit "Audience canonicalization" callout above the verifier snippet showing `https://MCP.Example.com/` → `https://mcp.example.com` and `https://your-server.example.com/v1/` → `https://your-server.example.com/v1`. Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — `iat` and `nbf` rows now show `now − 30s (clock-skew leeway)`. Added a "Clock-skew leeway" callout below the table explaining the backdating and recommending `clockTolerance: 30` (mirroring the agent-security README). Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — dropped "(enables replay detection)" from the `jti` row (now "Unique token ID") and added a "Replay protection (opt-in)" callout in the verifier section explaining that `jose.jwtVerify` does not deduplicate `jti`, with the suggested Redis SET pattern (TTL = `GATEWAY_JWT_TTL_SECONDS + 30`). Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — verifier snippet now guards a missing or non-string `x-gateway-auth` header, validates the `Bearer ` prefix, and sets `clockTolerance: 30`. Reworded the surrounding prose from "~5 lines" to "under 10 lines of meaningful code". Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — sentence now reads: "The `aud` claim is the **canonicalized form** of your upstream URL — host lowercased, all trailing slashes (including the root `/`) stripped — so configure `audience` with the canonicalized form (e.g. `https://MCP.Example.com/` becomes `https://mcp.example.com`)." The verifier snippet's `audience` line also has an inline canonical-form comment. Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — key-rotation callout rewritten to:

1. Qualify encryption-at-rest as conditional on `VAULT_ENABLED=true` + `AWS_KMS_KEY_ID` (with a recommendation to set `GATEWAY_JWT_REQUIRE_VAULT=true` in production to refuse the plaintext fallback)
2. State clearly that rotation is **manual** — the 90-day mark only emits a warn log + the `gateway_jwt_signing_key_age_seconds` metric — and show the `redis-cli DEL gateway:jwt:signing_key` + rolling-restart procedure

Pushed in 62447ff.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — configuration table now lists both vars:

- `GATEWAY_JWT_TTL_SECONDS` (default `300`) explicitly notes the `60 ≤ ttl ≤ 3600` bounds and that out-of-range values fail config validation at startup
- `GATEWAY_JWT_REQUIRE_VAULT` (default `false`) added with the production recommendation

Pushed in 62447ff.

zeevmoney (commented):
Automated Review - the PR looks clean, found nothing.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,139 @@ The derived role then determines which tools are allowed:
...
+    Note over GW: Startup: load shared Ed25519 key<br/>from Redis (or generate + store if none exists)
+
+    Client->>GW: POST /mcp (tool call)
+    GW->>GW: Auth + Permit check
+
+    rect rgb(240, 248, 255)
+        Note over GW: JWT signing (per-request)
+        GW->>GW: Is cached JWT still valid?
+        alt JWT expired or near expiry (within 30s)
+            GW->>GW: Sign fresh JWT<br/>{iss, sub, aud, exp, iat, jti, tenant}
```
zeevmoney: Sequence-diagram claim list omits `nbf`

The note enumerates the signed claims as `{iss, sub, aud, exp, iat, jti, tenant}` but omits `nbf`, which the JWT claims table below (the `nbf` row) and the implementation both include (`gateway/src/clients/gateway_jwt.rs` sets `nbf = now - NBF_LEEWAY_SECS`). Because the note explicitly lists each claim, the omission reads as if `nbf` isn't signed.

**Suggestion:** Add `nbf` to the list:

```
GW->>GW: Sign fresh JWT<br/>{iss, sub, aud, exp, iat, nbf, jti, tenant}
```

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -185,7 +186,7 @@ flowchart TD
 | Gateway | Consent Service (JWKS) | HTTP (port 3000) | JWT signature verification |
 | Gateway | Permit.io | HTTPS | Authorization checks (Cloud PDP) |
 | Consent Service | Permit.io | HTTPS | Policy sync during consent |
-| Gateway | Upstream MCP | HTTPS | Streamable HTTP with upstream OAuth tokens |
+| Gateway | Upstream MCP | HTTPS | Streamable HTTP with upstream OAuth tokens + signed `X-Gateway-Auth` JWT (see [Upstream Authentication](#upstream-authentication-gateway-jwt)) |
```
zeevmoney, line 189: Data Flow **diagram** edge not updated to match this table row

This row now mentions the signed `X-Gateway-Auth` JWT, but the Data Flow **diagram** edge a few lines above (the `GW -->|"Streamable HTTP +<br/>upstream OAuth tokens"| MCP` edge) was not updated, so the diagram and this table disagree.

(Copilot raised the same disagreement on this line; the actual fix is on that diagram edge.)

**Suggestion:** update the diagram edge to match:

```
GW -->|"Streamable HTTP + upstream OAuth tokens<br/>+ X-Gateway-Auth JWT"| MCP
```

zeevmoney (changes requested):
See comments.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,139 @@ The derived role then determines which tools are allowed:
...
+        GW->>GW: Is cached JWT still valid?
+        alt JWT expired or near expiry (within 30s)
+            GW->>GW: Sign fresh JWT<br/>{iss, sub, aud, exp, iat, jti, tenant}
+            GW->>GW: Cache JWT
+        else JWT still valid
+            GW->>GW: Reuse cached JWT
+        end
+    end
+
+    GW->>Up: POST (MCP request)<br/>Authorization: Bearer <upstream_oauth_token><br/>X-Gateway-Auth: Bearer <gateway_jwt>
```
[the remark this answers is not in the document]
dshoen619: Done — swapped the angle-bracket placeholders for `{upstream_oauth_token}` / `{gateway_jwt}` in the sequence diagram so they no longer read as HTML tags. Pushed in ca8ab6f.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,139 @@ The derived role then determines which tools are allowed:
...
+    participant JWKS as Gateway JWKS<br/>/.well-known/gateway-jwks.json
+    participant Up as Upstream MCP Server
+
+    Note over GW: Startup: load shared Ed25519 key<br/>from Redis (or generate + store if none exists)
+
+    Client->>GW: POST /mcp (tool call)
+    GW->>GW: Auth + Permit check
+
+    rect rgb(240, 248, 255)
+        Note over GW: JWT signing (per-request)
```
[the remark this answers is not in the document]
dshoen619: Done — reworded the note to `JWT signing (lazy — cached, re-signed near expiry)` so it no longer implies per-request signing. Pushed in ca8ab6f.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,139 @@ The derived role then determines which tools are allowed:
...
+
+| Claim    | Value                                                                                            | Purpose                           |
+| -------- | ------------------------------------------------------------------------------------------------ | --------------------------------- |
+| `iss`    | `agent-security-gateway`                                                                         | Identifies the issuer             |
+| `sub`    | Authenticated user ID                                                                            | Who the request is being made for |
+| `aud`    | Canonicalized upstream URL — host lowercased, all trailing slashes (including root `/`) stripped | Intended recipient                |
+| `exp`    | `now + TTL` (default 5 min)                                                                      | Token expiry                      |
+| `iat`    | `now − 30s` (clock-skew leeway)                                                                  | When the token was issued         |
+| `nbf`    | `now − 30s` (clock-skew leeway)                                                                  | Not valid before this time        |
```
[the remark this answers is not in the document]
dshoen619: Done — replaced the Unicode minus with an ASCII hyphen in the `iat`/`nbf` rows (`now - 30s`). Pushed in ca8ab6f.

docs/permit-mcp-gateway/architecture.mdx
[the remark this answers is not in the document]
dshoen619: Done — updated the Data Flow diagram edge to `Streamable HTTP + upstream OAuth tokens<br/>+ X-Gateway-Auth JWT`, so the diagram now matches the table row. Pushed in ca8ab6f.

docs/permit-mcp-gateway/architecture.mdx
```diff
@@ -440,6 +441,139 @@ The derived role then determines which tools are allowed:
...
+
+:::note Replay protection (opt-in)
+The `jti` claim is unique per token but `jose.jwtVerify` does not deduplicate it. If you need replay protection, track recently-seen `jti` values for at least the JWT TTL plus your clock-skew window — e.g. a Redis `SET` with `TTL = GATEWAY_JWT_TTL_SECONDS + 30` and reject any `jti` already in the set.
+:::
+
+:::info Key rotation
+The signing key is stored in Redis and shared across all gateway replicas, so JWTs remain valid across pod restarts and the JWKS endpoint returns the same key regardless of which replica serves it. When the token vault is enabled (`VAULT_ENABLED=true` + `AWS_KMS_KEY_ID`), the key is encrypted at rest using AES-256-GCM with AWS KMS envelope encryption; without vault it is stored as plaintext JSON (the gateway logs a warning). Set `GATEWAY_JWT_REQUIRE_VAULT=true` in production to refuse the plaintext fallback.
```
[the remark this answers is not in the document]
dshoen619: Leaving this as-is intentionally. The specific vault/KMS wording (`VAULT_ENABLED`, `AWS_KMS_KEY_ID`, AES-256-GCM, `GATEWAY_JWT_REQUIRE_VAULT`) was added at @zeevmoney's explicit request in an earlier review thread, because the unqualified "encrypted at rest" claim was misleading — encryption is conditional on the vault being enabled, and without it the key is stored as plaintext JSON. The flag is operationally important (it's the only way to prevent the plaintext fallback), so it's surfaced here rather than buried. Deferring to the human reviewer's accepted guidance over the original scope note.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — added `nbf` to the claim list: `{iss, sub, aud, exp, iat, nbf, jti, tenant}`. Pushed in ca8ab6f.

docs/permit-mcp-gateway/architecture.mdx
dshoen619: Done — updated the Data Flow diagram edge to `GW -->|"Streamable HTTP + upstream OAuth tokens<br/>+ X-Gateway-Auth JWT"| MCP`, matching your suggestion and the table row. Pushed in ca8ab6f.

zeevmoney (commented):
Automated Review - the PR looks clean, found nothing.

zeevmoney (approved):

PR #624, opened 2026-04-21
Add Cloud PDP benchmarks documentation

omer9564:
## Summary
- Adds a new **Cloud PDP Benchmarks** page under Concepts → PDP with RBAC and ReBAC performance data
- Includes latency percentiles (P50–P99), throughput, and results across concurrency levels (1 and 10) and ReBAC relationship depths (1–3 hops)
- Adds sidebar entry under "The Policy Decision Point (PDP)" section

## Test plan
- [ ] Verify the page renders correctly on the Netlify preview
- [ ] Confirm sidebar link appears under PDP section
- [ ] Review benchmark data accuracy


docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+| Max latency | 71.1 ms | 21.6 ms | 125.8 ms |
+| P50 | 12.0 ms | 10.8 ms | 16.4 ms |
+| P75 | 15.1 ms | 12.2 ms | 25.8 ms |
+| P90 | 20.5 ms | 15.0 ms | 36.8 ms |
+| P95 | 26.7 ms | 16.4 ms | 45.0 ms |
+| P99 | 48.6 ms | 18.9 ms | 73.9 ms |
+
+## Key Takeaways
+
+- **Sub-15 ms average latency** across all policy models at the P50 level, even under concurrent load.
```
zeevmoney: **[LOW]** Statistic mislabeled — "average latency" and "P50" are conflated

This takeaway reads "Sub-15 ms **average latency** ... at the **P50 level**", but average (mean) and P50 (median) are two different statistics. The reader can't tell which one the "Sub-15 ms" claim refers to. Both happen to be true against the tables (P50 Avg ranges 6.43–12.0 ms; Average-latency Avg ranges 7.80–14.5 ms), so pick one and state it cleanly.

**Suggestion:** Reword to a single statistic, e.g. "P50 latency stays under 15 ms across all policy models, even under concurrent load" (or "average latency under 15 ms ...").

(Copilot raised this on this line and it is still unresolved/unaddressed.)

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+- **Throughput scales linearly** — 10 concurrent requests yield ~10x throughput with minimal latency increase.
```
zeevmoney: **[MEDIUM]** "Throughput scales linearly" overclaims from two data points

The data has only two concurrency levels (1 and 10), so "scales linearly" is stronger than the measurements support — two points can't establish linearity, and the relationship isn't even exactly 10x (RBAC: 5.80 → 56.5 req/s ≈ 9.7x; ReBAC depth-1: 5.43 → 54.7 ≈ 10.1x). It is also an artifact of the paced load model (see the methodology note): with the PDP far from saturation, N paced clients trivially yield ~N× the per-client rate, which says nothing about the PDP's throughput ceiling. Soften to the observed claim.

**Suggestion:** "Throughput scales near-linearly with concurrency — 10 concurrent requests yield roughly 10x throughput with minimal latency increase."

(Copilot raised this on this line and it is still unresolved/unaddressed.)

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
+---
+sidebar_position: 3
+title: Cloud PDP Benchmarks
+---
+
+# Cloud PDP Benchmarks
+
+This page presents performance benchmarks for the **Cloud PDP**, measured across different policy models, concurrency levels, and relationship depths.
+
+All tests were run against a production Cloud PDP endpoint over a **10-minute window** per scenario. Latencies are round-trip times (request → response) over HTTPS.
```
zeevmoney: **[MEDIUM]** Benchmark methodology underspecified

For readers to interpret these numbers, the intro should state the measurement conditions: the date/period of the run, the Cloud PDP region and the client (load-generator) region/proximity, which endpoint(s) were exercised (e.g. `POST /allow`), the request/payload shape, and the policy size (number of roles/resources/relationship tuples).

The **load model** matters most. At single concurrency the reported `5.80 req/s` is far below what a back-to-back closed loop would yield for the stated latency — `1 / 8.17 ms ≈ 122 req/s`. Reaching only 5.80 req/s implies ~165 ms of think-time/pacing per request. That means the throughput figures reflect the **load generator's offered rate, not the PDP's capacity**, and the same holds at 10 concurrency (`56.5 req/s` vs the ~1,280 req/s the 7.80 ms latency could sustain). State this explicitly so the numbers aren't read as a throughput ceiling.

**Suggestion:** Add a short "Methodology" paragraph under the intro listing date, regions, endpoint, payload, policy size, and load model (open- vs closed-loop, plus any per-client request rate / think-time).

zeevmoney (commented):
This is an automated review. See comments.

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+| Max latency | 63.9 ms | 21.22 ms | 258.5 ms |
+| P50 | 9.5 ms | 8.57 ms | 14.5 ms |
+| P75 | 11.7 ms | 9.40 ms | 22.8 ms |
+| P90 | 15.9 ms | 10.81 ms | 33.5 ms |
+| P95 | 20.1 ms | 12.05 ms | 43.0 ms |
+| P99 | 40.0 ms | 13.43 ms | 153.2 ms |
+
+### Depth 2 — 10 Concurrent Requests
+
+*Drive → Folder → File*
```
zeevmoney, line 87: **[MEDIUM]** Benchmark scenario uses resources not present in the cloud-pdp ReBAC fixtures

A thorough search of the `cloud-pdp` repo found no benchmark script or results file that produces these numbers, and the ReBAC fixtures (`connector-service/src/sequin/faker.rs`, `connector-service/src/sequin/records.rs`) only model `Folder` and `File`. This Depth 2 scenario (`Drive → Folder → File`) and the Depth 3 scenario on line 94 (`Drive → Folder → File → Comment`) reference `Drive` and `Comment` resource types that don't exist in those fixtures.

The numbers may come from an internal/ad-hoc run, but as written the scenario model can't be verified against the source, and the page is not reproducible.

**Suggestion:** Confirm the actual resource schema used for the run with the author, and link the benchmark tooling (or commit it) so the scenarios and numbers are traceable.

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+| Average latency | 11.3 ms | 9.07 ms | 18.2 ms |
+| Min latency | 6.5 ms | 5.90 ms | 7.8 ms |
+| Max latency | 63.9 ms | 21.22 ms | 258.5 ms |
+| P50 | 9.5 ms | 8.57 ms | 14.5 ms |
+| P75 | 11.7 ms | 9.40 ms | 22.8 ms |
+| P90 | 15.9 ms | 10.81 ms | 33.5 ms |
+| P95 | 20.1 ms | 12.05 ms | 43.0 ms |
+| P99 | 40.0 ms | 13.43 ms | 153.2 ms |
+
+### Depth 2 — 10 Concurrent Requests
```
zeevmoney, line 85: **[LOW]** Asymmetric concurrency coverage across ReBAC depths

RBAC and ReBAC Depth 1 each present both a single-concurrency and a 10-concurrent table, but Depth 2 and Depth 3 present only the 10-concurrent table. The inconsistency makes the depth comparison harder to read and leaves a gap for the single-concurrency case at higher depths.

**Suggestion:** Either add the single-concurrency tables for Depth 2 and Depth 3, or add a one-line note explaining why only the 10-concurrent results are shown for those depths.

zeevmoney (approved):
## Approved ✅ — with comments to address before merge

Thanks @omer9564 — the page is a solid addition and the data tables are internally consistent. Approving so this isn't blocked, but please look over the review comments first.

**Please see the inline comments, address them, and then merge.** A few are worth a quick pass:

- **[MEDIUM]** Add a short methodology note (line 10) — date, regions, endpoint, payload, policy size, and the load model. The throughput figures reflect the load generator's offered rate (~paced), not the PDP's capacity, so it's worth saying so explicitly.
- **[MEDIUM]** The Depth 2/3 scenarios (lines 76–94) reference `Drive` and `Comment` resources that don't appear in the `cloud-pdp` ReBAC fixtures, and I couldn't trace these numbers to any benchmark script in that repo — please confirm the schema/source and ideally link the tooling.
- **[MEDIUM]** Soften "Throughput scales linearly" (line 111) and **[LOW]** fix the mean-vs-P50 wording (line 110).
- **[LOW]** Depth 2/3 are missing the single-concurrency tables that RBAC and Depth 1 have (line 76).

None of these block the doc from going live — they're accuracy/clarity improvements. Fix them up and merge when you're happy. 🚀

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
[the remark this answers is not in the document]
omer9564: Fixed in be587091 — reworded the takeaway to a single statistic: "P50 latency stays under 15 ms across all policy models, even under concurrent load." Thanks!

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+| P75 | 15.1 ms | 12.2 ms | 25.8 ms |
+| P90 | 20.5 ms | 15.0 ms | 36.8 ms |
+| P95 | 26.7 ms | 16.4 ms | 45.0 ms |
+| P99 | 48.6 ms | 18.9 ms | 73.9 ms |
+
+## Key Takeaways
+
+- **Sub-15 ms average latency** across all policy models at the P50 level, even under concurrent load.
+- **Throughput scales linearly** — 10 concurrent requests yield ~10x throughput with minimal latency increase.
+- **ReBAC depth has modest impact** — adding relationship depth (1 → 3 hops) increases average latency by only ~3 ms under the same concurrency.
```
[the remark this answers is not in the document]
omer9564: Fixed in be587091 — softened to "Throughput scales near-linearly with concurrency — 10 concurrent requests yield roughly 10x throughput with minimal latency increase." Thanks!

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
```diff
@@ -0,0 +1,120 @@
...
+sidebar_position: 3
+title: Cloud PDP Benchmarks
+---
+
+# Cloud PDP Benchmarks
+
+This page presents performance benchmarks for the **Cloud PDP**, measured across different policy models, concurrency levels, and relationship depths.
+
+All tests were run against a production Cloud PDP endpoint over a **10-minute window** per scenario. Latencies are round-trip times (request → response) over HTTPS.
+
```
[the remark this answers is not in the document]
omer9564: Fixed in be587091 — added a Methodology section (date, region, measurement source) and clarified what the Avg/Min/Max columns represent. Also corrected the latency description: these are the PDP's server-side processing times, not network round-trips.

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
omer9564: Fixed in be587091 — reworded to a single statistic: "P50 latency stays under 15 ms across all policy models, even under concurrent load." Thanks!

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
omer9564: Fixed in be587091 — adopted your wording: "Throughput scales near-linearly with concurrency — 10 concurrent requests yield roughly 10x throughput with minimal latency increase." Thanks!

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
omer9564: Fixed in be587091 — added a Methodology section: October 2025, AWS us-east-2. Important clarification: the numbers come from the PDP's server-side metrics and exclude client↔PDP network latency — the previous "round-trip over HTTPS" wording was incorrect and has been fixed. I also noted that throughput is the observed request rate during the run (not a saturation ceiling) and clarified what the Avg/Min/Max columns represent.

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
omer9564: Thanks for the careful check. These numbers come from an internal/ad-hoc benchmark run measured from the PDP's server-side metrics, not from the cloud-pdp repo's ReBAC fixtures — which is why Drive/Comment don't appear there. The Depth 2/3 scenarios (Drive → Folder → File → Comment) were set up specifically for that run. I've added a Methodology section in be587091 noting the measurement source; the run tooling isn't committed to cloud-pdp.

docs/concepts/pdp/cloud-pdp-benchmarks.mdx
omer9564: Fixed in be587091 — added a note explaining that single-concurrency results are shown for RBAC and Depth 1 as a baseline, while Depth 2 and Depth 3 report only the 10-concurrent (realistic concurrent-load) case. Thanks!

PR #638, opened 2026-06-12
PER-15119: remove deprecated pdp-api.permit.io host from docs

EliMoshkovich:
## What

Removes all references to the shared public PDP host `pdp-api.permit.io` from the docs. That instance is a legacy/decommissioned PDP (no longer in use), so any docs pointing at it were either broken (the embedded online Redoc) or steering users at a shared host they shouldn't be calling.

## Changes

| File | Change |
|------|--------|
| `docs/api/pdp-api-reference.mdx` | Dropped the dead online Redoc `<iframe>` embed (`pdp-api.permit.io/redoc`); kept and tidied the local-PDP OpenAPI/Redoc guidance. |
| `docs/sdk/nodejs/all-tenants.mdx` | Redoc deep-link → internal [`/api/pdp-api-reference`](/api/pdp-api-reference); curl example host → `http://localhost:7766`. |
| `docs/how-to/enforce-permissions/url-mapping/url-mapping-check.mdx` | Redoc deep-link → internal `/api/pdp-api-reference`. |
| `src/sdks/walkthroughs/authz-queries/getAuthorizedUsers/example.bash` | curl host → `http://localhost:7766`. |
| `src/sdks/walkthroughs/authz-queries/getUserPermissions/example.bash` | curl host → `http://localhost:7766`. |

## Notes

- `http://localhost:7766` is the established local-PDP convention across the docs for these authz endpoints (`/allowed`, `/user-permissions`, `/authorized_users`), so the curl examples now match the rest of the docs.
- Left untouched: the `permit-pdp-api-key` GCP secret name in `gcp-cloud-run.mdx` — that's a secret name, not the host.
- The `api/pdp-api-reference` page (and its sidebar entry) is kept, just reframed around the local PDP, so no inbound links or sidebar entries break.
- Open question for reviewers: if we want to keep an *online* embedded Redoc, we could repoint it at `cloudpdp.api.permit.io/redoc` (the current hosted PDP) instead — left out here since the ask was to remove the dead host, not introduce a new embed.

### DIFF
--- a/docs/api/pdp-api-reference.mdx
+++ b/docs/api/pdp-api-reference.mdx
@@ -9,4 +9,4 @@ title: PDP API Reference
 
 Just like the Permit.io cloud service, the PDP exposes an OpenAPI spec which can be used to explore the API and develop your own SDKs.
 
-Assuming you're running your PDP at `localhost:7000` you can access the spec at `http://localhost:7000/openapi.json` and view the docs at `http://localhost:7000/redoc`.
+Assuming you're running your PDP at `localhost:7766` you can access the spec at `http://localhost:7766/openapi.json` and view the docs at `http://localhost:7766/redoc`.
--- a/docs/api/pdp-api-reference.mdx
+++ b/docs/api/pdp-api-reference.mdx
@@ -5,23 +5,8 @@ title: PDP API Reference
 
 # PDP API Reference
 
-## PDP/SDK OpenAPI spec - overview and evaluation purpose of online version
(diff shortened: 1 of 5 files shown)


zeevmoney (commented):
Automated review — every issue this review found is already covered by existing review comments (e.g. GitHub Copilot or a prior review), so no new comments were posted. Please address those existing comments.

zeevmoney (approved):
See co-pilot comment, approved.

PR #649, opened 2026-09-07
Add New Edge PDP documentation

dshoen619:
Adds documentation for the new PDP runtime (`permitio/pdp-v3`), documented as the **New Edge PDP** — an **additional** deployment option. The existing `permitio/pdp-v2` is referred to throughout as the **container PDP** and is framed as fully supported; the parity table reads as "what the new one covers so far", never as removals.

Four pages, wired into `sidebars.js` as a collapsible **New Edge PDP** sub-category under **The Policy Decision Point (PDP)**:

| Page | Contents |
| --- | --- |
| **Overview** | what it is, why, architecture diagram, choosing between the two |
| **How It Works** | sync, consistency, propagation latency, cold start, scale, resource footprint |
| **Feature Parity** | capability comparison against the container PDP |
| **Deployment** | requirements, observability, security model |

## Accuracy

Every endpoint, port, and behaviour was verified against the `cloud-pdp` source rather than that repo's README or staging runbook, both of which are stale in several places.

Two errors caught during review, both wrong in *both* columns:

- **Partial-evaluation data filtering** was scored as a container-PDP feature and a New Edge PDP gap. There is no compile endpoint in `horizon/` or `pdp-server/` — neither runtime has it. Row removed; graph-based filtering (which both support) is covered by the user-permissions and authorized-users rows.
- **Envoy ext-authz** was listed as a New Edge PDP gap. The container PDP does not have it either. Removed.

## Known gaps, documented as such

- **ABAC** is not evaluated — a check against an ABAC policy returns a deny. Use the container PDP.
- **Absolute resource footprint is not reduced.** CI pins a 4 GiB limit and notes that 500 MiB OOM-kills. The documented win is architectural: memory stops scaling with data set size.
- **No latency or throughput numbers** exist in the repo, so the pages make the structural argument and say explicitly that figures are not published.

## OpenTelemetry

Not implemented in `edge-pdp` today — it never calls `init_observability`, has no `opentelemetry` dependency, reads no `OTEL_*` variables, and serves no `/metrics`. Both in-repo OTEL branches predate the crate.

Per review, it **is on the roadmap**, so the parity table marks it 🚧 for the New Edge PDP and ❌ for the container PDP, which has Prometheus `/metrics` but no OTLP export. No timeline is stated anywhere.

## Configuration

Deliberately **not documented yet** — the environment variable surface is still settling, and publishing a reference now would ship something known to go stale. The operational requirements that readers act on (persistent storage, ≥40s `terminationGracePeriodSeconds`, probe targets, memory sizing, ports) live on the Deployment page without pinning variable names.

## Follow-up found along the way (not fixed here)

`docs/concepts/pdp/cloud-pdp-capabilities.mdx:59-63` lists the AuthZen paths as `/v1/access/evaluation`, `/v1/subjects`, `/v1/resources`, `/v1/actions`. The actual routes are `/access/v1/evaluation`, `/access/v1/evaluations`, `/access/v1/search/{subject,resource,action}`. That same page's rate-limit table at lines 88-93 already uses the correct form, so it contradicts itself.

## Verification

- `npm run redirect-lint` — clean
- `npx docusaurus build` — succeeds, no broken links; cross-page anchors confirmed to resolve
- Architecture diagram is a static SVG extracted from the source diagram, so the image-zoom plugin applies — click-to-zoom verified in a browser, in both light and dark palettes
- Every table on all four pages measured to fit the content column (an earlier four-column parity table was being clipped, not scrolled)

https://claude.ai/code/session_011ghyfrjy3SGKPV8ruaXE8W


docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,436 @@
...
+| [Check](/how-to/enforce-permissions/check) | `POST /allowed` |
+| [Bulk check](/how-to/enforce-permissions/bulk-check) | `POST /allowed/bulk` |
+| [User permissions](/how-to/enforce-permissions/user-permissions) | `POST /user-permissions` |
+| [Authorized users](/how-to/enforce-permissions/authorized-users) | `POST /authorized_users` |
+| AuthZen access evaluation | `POST /access/v1/evaluation` |
+| AuthZen bulk evaluations | `POST /access/v1/evaluations` |
+| AuthZen subject / resource / action search | `POST /access/v1/search/{subject,resource,action}` |
+| AuthZen discovery | `GET /.well-known/authzen-configuration` |
+| Decision logs to Permit | — |
+| Health and readiness | `GET /health`, `GET /healthy` |
```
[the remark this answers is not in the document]
dshoen619: Good catch, though the cause is different from what's suggested here — and the real version is worse, so thanks for surfacing it.

Both endpoints are real. They're on different listeners:

- **Authorization port `7000`** (`cloud-pdp-core/src/routes/mod.rs:19,31`) — serves `/health` and `/healthy`. The `/healthy` spelling is a deliberate alias so tooling written against the existing Edge PDP keeps working.
- **Management port `7001`** (`edge-pdp/src/servers/management.rs:67,75,76`) — serves `/health`, `/health/ready` and `/health/detail`.

So rather than remove either, I've made the split explicit and documented the two ways it bites in practice:

1. A readiness probe on `:7000/health/ready` gets a `404` and the pod never becomes ready.
2. Port `7000` doesn't open until the PDP is ready, so *any* probe against it fails for the entire cold start — which can run to minutes. A liveness probe there would kill the container before it ever finished booting.

Fixed in 958a620.

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,436 @@
...
+
+| Variable | Description |
+| --- | --- |
+| `PDP_API_KEY` | Your environment-level Permit API key. This single value binds the container to one environment, authenticates it to the control plane, and is the bearer token the authorization API accepts. |
+| `PDP_CONTROL_PLANE` | The control-plane endpoint. **See the warning below — this is not the same kind of value as on `pdp-v2`.** |
+
+:::warning `PDP_CONTROL_PLANE` means something different here
+On the existing Edge PDP, `PDP_CONTROL_PLANE` is an **HTTPS API base URL** (`https://api.permit.io`).
+
+On the New Edge PDP it is the **event-stream endpoint**, and it must be a bare host and port with **no path and no trailing slash** — for example `wss://control-plane.permit.io:443`. Copying the `pdp-v2` value across will not work.
```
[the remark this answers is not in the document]
dshoen619: Agreed — "bare host and port" contradicted the `wss://` in the example. Reworded to say scheme + host + explicit port, with no path and no trailing slash, and called out that the port is not optional. Fixed in 958a620.

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,407 @@
...
+
+The New Edge PDP changes all three. Data lives in an **embedded on-disk database**, updates **arrive already containing the change**, and the control plane **durably retains** each PDP's unacknowledged changes until that specific PDP confirms it applied them.
+
+---
+
+## Architecture
+
+Everything runs inside one container. The only externally exposed listeners are the authorization API and the health endpoint.
+
+![Architecture of the New Edge PDP: your app and orchestrator on the left, the single Edge PDP container holding the authorization API, OPA, the health server, the query and bundle loopbacks, the embedded database, the change-ingest worker and the NATS leaf, with Permit.io and your policy store on the right](/images/pdp/new-edge-pdp-architecture.svg)
```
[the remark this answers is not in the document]
dshoen619: Correct, the alt text said "Edge PDP container" while the page consistently says New Edge PDP. Fixed in 958a620.

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+The boundary between the two is exact: the snapshot records the precise stream position it was cut at, and the PDP resumes the change stream from the very next change. Nothing is applied twice in a way that matters, and nothing is skipped.
+
+A checksum mismatch, a missing chunk, or an empty snapshot is **fail-closed** — the PDP refuses to come up on a partial data set rather than serving decisions against one.
+
+:::caution Cold start is not fast — plan probes accordingly
+A cold start transfers and loads the environment's entire data set. Permit's own CI observes **84 seconds to roughly 4 minutes** for this on test environments, and larger environments will take longer.
+
+The health endpoint answers throughout, so **set a generous startup probe** and let readiness — not liveness — gate traffic. The authorization port does not open until the PDP is ready, so there is no window in which it accepts a request it cannot answer correctly.
+:::
```
omer9564: where did you take this data from ? the new edge pdp has faster coldstart then the old one

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+3. **Per-transaction atomicity.** All operations within a transaction are applied together, so a decision never observes half of a multi-part change.
+
+Together these give **convergence by construction**: two PDPs in the same environment that have seen the same set of changes hold the same state, regardless of the order or timing in which they received them, and re-delivering a change that was already applied is a no-op.
+
+:::note What this does and does not guarantee
+The consistency model is **eventual, with a convergent merge**. It guarantees that PDPs converge, that no update is lost, and that replays are safe.
+
+It does **not** provide read-your-own-writes. There is no local write API — the New Edge PDP is read-only with respect to your data, and a fact written through Permit's API becomes visible once it propagates. If you need write-then-immediately-read semantics, use the existing Edge PDP's [local facts uploader](/how-to/manage-data/local-facts-uploader).
```
omer9564: this might be a feature we'll add in the future, but don't commit to we adding it at some point or provide a timeline

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+Its capability envelope is that of the **managed Cloud PDP, deployed in your own network** — not a port of the `pdp-v2` container.
+
+### Supported
+
+| Capability | Endpoint |
+| --- | --- |
+| **RBAC** | — |
+| **ReBAC**, including role derivation | — |
+| **Multi-tenancy** | — |
+| [Check](/how-to/enforce-permissions/check) | `POST /allowed` |
```
omer9564: these are basically a part of each check, it should be some sort of nested table of 
Check
 - RBAC 
 - ReBAC ...

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+| [User permissions](/how-to/enforce-permissions/user-permissions) | `POST /user-permissions` |
+| [Authorized users](/how-to/enforce-permissions/authorized-users) | `POST /authorized_users` |
+| AuthZen access evaluation | `POST /access/v1/evaluation` |
+| AuthZen bulk evaluations | `POST /access/v1/evaluations` |
+| AuthZen subject / resource / action search | `POST /access/v1/search/{subject,resource,action}` |
+| AuthZen discovery | `GET /.well-known/authzen-configuration` |
+| Decision logs to Permit | — |
+| Health check on the authorization port | `GET /health`, `GET /healthy` (aliases for the same handler) |
+
+### Not supported
```
omer9564: I think it's better to create a matrix of feature compatibility and roadmap

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+### Decision logs
+
+Decision logs ship from the OPA child directly to Permit over HTTPS, batched, with a bounded in-memory buffer. They require **both** that decision logging is enabled for your environment in Permit **and** that `EDGE_DECISION_LOG_OPT_OUT` is not set — the local setting can only turn logging off, never on.
+
+:::caution OpenTelemetry is not available yet
+The New Edge PDP **does not currently export OpenTelemetry traces, metrics, or logs**, and does not read `OTEL_*` environment variables. It also does not serve a `/metrics` endpoint.
+
+Observability today is structured stdout logs plus the health endpoints described above. If OTLP export is a requirement for your deployment, [let us know](mailto:support@permit.io) — it is a known gap rather than a design decision.
+:::
```
omer9564: this is a part of the feature parity table ( old pdp doesn't have it, new is in the roadmap )

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+---
+
+## Configuration
+
+### Required
+
+| Variable | Description |
+| --- | --- |
+| `PDP_API_KEY` | Your environment-level Permit API key. This single value binds the container to one environment, authenticates it to the control plane, and is the bearer token the authorization API accepts. |
+| `PDP_CONTROL_PLANE` | The control-plane endpoint. **See the warning below — this is not the same kind of value as on `pdp-v2`.** |
```
omer9564: there is no need to configure this anymore, the apikey of the v3 automatically routes the PDP to the correct plane

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,416 @@
...
+:::
+
+### Storage and ports
+
+| Variable | Default | Description |
+| --- | --- | --- |
+| `EDGE_DB_PATH` | `/var/lib/edge-pdp/db` | Embedded database directory. **Must be on a persistent volume.** |
+| `EDGE_DATA_DIR` | `/var/lib/edge-pdp` | Event store and generated credentials. **Must be on a persistent volume.** |
+| `EDGE_PUBLIC_PORT` | `7000` | Authorization API. The only port your services need. |
+| `EDGE_HEALTH_PORT` | `7001` | Health and readiness. |
```
omer9564: fyi - some env vars might change here in the future

docs/concepts/pdp/new-edge-pdp.mdx
omer9564, line 1: It might be beneficial to split some parts here to a different page, e.g configuration as we already have a configuration page for the old one

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: From `.github/workflows/run_edge_pdp_tester.yml:146` in cloud-pdp — `START_TIMEOUT: "900"  # cold boot 84s-4min draining the snapshot`, alongside the job comment at line 60 saying "cold boot alone can take 4 min per PDP". So it was a CI observation, not a comparison against v2 — which is exactly the problem, since quoted with no baseline it reads as a criticism.

Rewritten. It now says bulk-loading pre-built database files is substantially faster than v2's cold start, which applies its data set through the normal write path, and the CI figure is gone. I kept the startup-probe guidance, since that's the part readers actually act on.

If there's a published v2-vs-v3 cold-start number I can cite, I'll add it — that would be a much stronger claim than the qualitative one.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Removed — good catch, and it would have been an actively harmful instruction.

Confirmed in the code too: `edge-pdp/src/config.rs:543` describes `PDP_CONTROL_PLANE` as an "operator ESCAPE HATCH, not the normal path", with the bundle's own `nats_url` winning by default. The page now says the API key carries the routing and there is no endpoint URL to configure.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Done — RBAC, ReBAC and multi-tenancy are now nested under **Check** with a `└` indent rather than sitting as peers of it, and ABAC joins them as the one that is not supported. That reads much better; they were misleading as top-level rows.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Done. The supported/not-supported split is replaced by a single compatibility matrix scoring each capability against both runtimes, with a ✅ / 🚧 roadmap / ❌ key, grouped into Permission checks, Standards, Policy and data, and Operations.

Building it surfaced two errors in my original lists, both now fixed:

- I had **Envoy ext-authz** as a New Edge PDP gap, but v2 doesn't have it either — dropped rather than shown as a regression.
- **Prometheus `/metrics`** and **OTLP export** are now separate rows. v2 has the former (`horizon/config.py:250`) but not the latter, so collapsing them would have got both columns wrong.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Moved into the compatibility matrix under **Operations** — OTLP export is `❌` for the Edge PDP and `🚧` roadmap for the New Edge PDP, with Prometheus `/metrics` as its own row since v2 does have that one. The Observability section keeps a short factual note pointing at the matrix, without the "known gap" phrasing.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Tightened. Removed the phrasings that read as commitments — "known gap rather than a design decision", "not available yet" — so the only forward-looking signal on the page is the 🚧 marker in the matrix, with no timeline attached anywhere. Read-your-own-writes is now stated as current behaviour only, pointing at the local facts uploader as the option that exists today.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Noted — the new configuration page carries an early-access callout saying variables and defaults may change before GA, and advising readers to pin an image tag and re-check on upgrade. Happy to mark specific variables as unstable if you know which ones are most likely to move.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Done — configuration is now its own page at `/concepts/pdp/new-edge-pdp-configuration`, mirroring the existing `/concepts/pdp/configuration`, and added to the sidebar under the PDP section. The main page keeps a short pointer plus the one variable that actually matters on day one (`PDP_API_KEY`). That also pulled the storage-engine tuning table off the main page. Main page is down from ~436 to ~400 lines with the reference material split out.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Followed up on this — rather than caveat a reference that's likely to go stale, we've pulled the environment variable reference out entirely for now. The configuration page is gone and the section on the main page keeps only `PDP_API_KEY`, pointing at support for anything else.

The operational guidance survives without the variable names: persistent storage is still mandatory, `terminationGracePeriodSeconds` is still stated as ≥40s (as a combined shutdown budget rather than naming the two vars), and the warning that turning up child log levels exposes authorization data is still there.

Easy to add back once the surface settles.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: I removed the page for now

omer9564, line 1: would rephrase the old PDP from the term Edge PDP as it is mostly confusing and unused term until now

omer9564: the configuration page does not appear on the sidebar + this page is still super long,  and need some more splitting

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,381 @@
...
+---
+
+## Choosing between the two
+
+| Choose the **existing Edge PDP** when you need | Choose the **New Edge PDP** when you need |
+| --- | --- |
+| ABAC, condition sets, user or resource sets | The largest data sets, without sharding |
+| Policy-as-Code / custom Rego bundles | Bounded, tunable memory rather than memory that scales with data |
+| Read-your-own-writes / local facts | Decisions that never depend on reaching Permit's API |
+| Data filtering via partial evaluation | Sync that survives long disconnections without a re-fetch |
```
omer9564: this is not a feature even on the old one, remove it

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,381 @@
...
+
+## Feature compatibility and roadmap
+
+The New Edge PDP is **wire-compatible on the endpoints it implements** — same paths, same request and response shapes, so existing SDK integrations that use those endpoints work unchanged. But it implements a **subset** of the existing Edge PDP's surface.
+
+Its capability envelope is that of the **managed Cloud PDP, deployed in your own network** — not a port of the `pdp-v2` container.
+
+**✅ Supported · 🚧 On the roadmap · ❌ Not supported**
+
+### Permission checks
```
omer9564: I imagine the whole thing as a big table of feature parity not splitted

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,381 @@
...
+
+Decision logs ship from the OPA child directly to Permit over HTTPS, batched, with a bounded in-memory buffer. They require **both** that decision logging is enabled for your environment in Permit **and** that the PDP's local opt-out is not set — the local setting can only turn logging off, never on.
+
+:::note OpenTelemetry and `/metrics`
+The New Edge PDP does not currently export OpenTelemetry traces, metrics, or logs, does not read `OTEL_*` environment variables, and does not serve a `/metrics` endpoint. OTLP export is on the roadmap — see [Feature compatibility](#feature-compatibility-and-roadmap).
+
+Observability today is structured stdout logs plus the health endpoints described above. If OTLP export matters for your deployment, [let us know](mailto:support@permit.io).
+:::
+
+---
```
omer9564: this doesn't feel like the right place to put this on

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Agreed — "Edge PDP" for v2 is gone from all these pages. It is now the **container PDP** (`permitio/pdp-v2`), which matches the wording already used in `cloud-pdp-capabilities.mdx`. "New Edge PDP" stays as the name for v3, per the naming decision on this PR.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Done — one table now, not four split by theme, with the check sub-capabilities (RBAC / ReBAC / multi-tenancy / ABAC) nested under **Check**.

Merging it did surface a rendering bug: at four columns the table was 732px in a 640px content column with `overflow-x: visible`, so the New Edge PDP column was silently clipped rather than scrollable. Folding the endpoint into the capability cell drops it to three columns; verified every table on all four pages now fits.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Removed — and you're right in both columns, so this was worse than a stray row.

Checked the v2 source: there's no compile/partial-evaluation endpoint in `horizon/` or `pdp-server/`. So scoring partial evaluation as `✅` for the container PDP and `❌` for the New Edge PDP was wrong twice over. The row is gone from the parity table and from "Choosing between the two".

Graph-based filtering — which both PDPs do support, via `/user-permissions` and `/authorized_users` — is already covered by those rows, and there's now a short note under the table pointing at the Data Filtering guide so readers arriving from there aren't left wondering.

Same reasoning retired the leftover ext-authz mention: v2 doesn't have that either.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Removed. It was duplicating the parity table, which is where OTLP now lives (`❌` container PDP, `🚧` New Edge PDP). The observability section just covers what exists today — health endpoints, structured logs, decision logs.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Both handled.

**Sidebar** — the configuration page has since been removed entirely (it was pulled after your env-var comment, since the surface is still moving). The four pages that exist now are all wired in, as a collapsible **New Edge PDP** sub-category under the PDP section with the overview as its landing page. Verified in a built site: the breadcrumb renders `The Policy Decision Point (PDP) › New Edge PDP › Feature Parity`.

**Length** — split from one ~380-line page into four:

- **Overview** — what it is, why, architecture, choosing between the two
- **How It Works** — sync, consistency, latency, cold start, scale, footprint
- **Feature Parity** — the comparison table
- **Deployment** — requirements, observability, security model

Happy to split further if any of them still reads long — How It Works is the biggest at ~155 lines.

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,396 @@
...
+title: New Edge PDP
+sidebar_label: New Edge PDP
+description: A single-binary PDP with an embedded database that keeps a local copy of your policy and authorization data in sync over a durable event stream — no network in the decision path.
```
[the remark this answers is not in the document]
dshoen619: Fair — it does contradict the architecture table two sections down, which lists OPA and the NATS leaf as supervised child processes.

I went a different way than "single-container", though: the container PDP is also a single container, so that phrasing would trade a contradiction for a claim that distinguishes nothing. Both frontmatter and the opening line now lead with what actually is distinctive — an embedded on-disk database, kept in sync over a durable event stream, no network in the decision path. The precise in-process-vs-child breakdown stays in the architecture table, which was already correct.

While fixing it I also dropped "fewer moving parts — one binary" from the comparison table. Both runtimes run three processes (Python API server + OPAL client + OPA, against Rust binary + NATS leaf + OPA), so that was simply false. It now reads "a compiled runtime, with no Python or OPAL layer".

Fixed in 984c38a.

docs/concepts/pdp/new-edge-pdp-feature-parity.mdx
```diff
@@ -0,0 +1,61 @@
...
+| **Policy-as-Code / custom Rego bundles** | ✅ | ❌ |
+| **Local facts / read-your-own-writes** | ✅ | ❌ |
+| [**Local enforcement APIs**](/how-to/enforce-permissions/list-role-assignments) — `/local/*` | ✅ | ❌ |
+| **Custom data sources** | ✅ | ❌ |
+| **Kong and nginx integration** | ✅ | ❌ |
+| [**Offline / air-gapped first boot**](/how-to/deploy/offline-mode) | ✅ | ❌ |
+| **Decision logs to Permit** | ✅ | ✅ |
+| **Health and readiness endpoints** | ✅ | ✅ |
+| **Prometheus metrics** — `GET /metrics` | ✅ | ❌ |
+| **OpenTelemetry (OTLP traces, metrics, logs)** | ❌ | 🚧 |
```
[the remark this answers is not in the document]
dshoen619: Good catch on the inconsistency — but the fix belongs on the PR description, not the table.

The description was written from a code investigation that established OTLP is **not implemented** (no `opentelemetry` dependency, `init_observability` never called, no `OTEL_*` read, no `/metrics`) and that no branch is in flight. That is evidence about the current tree, not about intent — "nothing in flight" is not the same as "not planned".

A maintainer then reviewed and stated it is on the roadmap, which is the authoritative source for that question. Hence 🚧 for the New Edge PDP and ❌ for the container PDP, which has Prometheus `/metrics` but no OTLP export.

The PR description was stale against that and has now been updated to say both things: not implemented today, on the roadmap per review. No timeline is stated on the docs pages or in the table.

docs/concepts/pdp/new-edge-pdp.mdx
omer9564: we should have configuration page, it should be marked as beta and open for changes but it is something we need

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,99 @@
...
+It is currently in early access and enabled per account. [Contact us](mailto:support@permit.io) to have it turned on.
+:::
+
+---
+
+## Why a new runtime
+
+The container PDP keeps your authorization data inside OPA's in-memory document, and refreshes it by reacting to a notification and then **fetching the changed data back from Permit's API**. That design is proven, but it has three structural ceilings:
+
+- **Memory scales with your data.** Because the fact graph lives in memory, a large environment needs a large PDP — which is why sharding exists for the biggest data sets.
```
omer9564: sharding is deprecated and should be removed, please remove it as agents keep referencing it 
( remove both from docs and code but at least a ticket at first and remove the reference here )

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+For how data reaches that database and what consistency it gives you, see [How It Works](./new-edge-pdp-how-it-works).
+
+---
+
+## Choosing between the two
+
+| Choose the **container PDP** when you need | Choose the **New Edge PDP** when you need |
+| --- | --- |
+| ABAC, condition sets, user or resource sets | The largest data sets, without sharding |
```
omer9564: sharding is deprecated and should be removed, please remove it as agents keep referencing it 
( remove both from docs and code but at least a ticket at first and remove the reference here )

docs/concepts/pdp/new-edge-pdp.mdx
omer9564: I'd say referencing [this](https://www.openpolicyagent.org/docs/policy-performance#:~:text=Memory%20usage%20scales%20with%20the%20size%20of%20the%20policy%20(i.e.%2C%20Rego)%20and%20data%20(e.g.%2C%20JSON)%20that%20you%20load%20into%20OPA.%20Raw%20JSON%20data%20loaded%20into%20OPA%20uses%20approximately%2020x%20more%20memory%20compared%20to%20the%20same%20data%20stored%20in%20a%20compact%2C%20serialized%20format%20(e.g.%2C%20on%20disk)) sentence from OPA docs is a good idea here

> Memory usage scales with the size of the policy (i.e., Rego) and data (e.g., JSON) that you load into OPA. Raw JSON data loaded into OPA uses approximately 20x more memory compared to the same data stored in a compact, serialized format (e.g., on disk)

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+:::note Availability
+The New Edge PDP is a **new deployment option**, not a replacement. The [container PDP](/concepts/pdp/overview#run-an-edge-pdp-with-docker) (`permitio/pdp-v2`) remains fully supported and continues to be the right choice for the capabilities listed under [Feature Parity](./new-edge-pdp-feature-parity).
+
+It is currently in early access and enabled per account. [Contact us](mailto:support@permit.io) to have it turned on.
+:::
+
+---
+
+## Why a new runtime
```
omer9564, line 22: add here something about that opa data system is a single json document that isn't and can't be optimized for actual graph traversal or complex granular permissions query requirements

docs/concepts/pdp/new-edge-pdp.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+- **Memory scales with your data.** Because the fact graph lives in memory, a large environment needs a large PDP — which is why sharding exists for the biggest data sets.
+- **Every update costs a round trip.** A notification tells the PDP *that* something changed; a second call is needed to find out *what*.
+- **Catching up after a disconnect is a re-fetch.** There is no server-side record of precisely what a given PDP has and hasn't seen.
+
+The New Edge PDP changes all three. Data lives in an **embedded on-disk database**, updates **arrive already containing the change**, and the control plane **durably retains** each PDP's unacknowledged changes until that specific PDP confirms it applied them.
+
+---
+
+## Architecture
```
omer9564: i'd say architecture needs its own subpage explaining the architecture, which part is responsible of what and stuff like that while the "root" page stays thin about the general idea, why it was established, what it allows and stuff like that

dshoen619: Done — Architecture is now its own subpage. It carries the diagram, the per-component responsibilities, a port table, and the request path, plus a new section on how the read path and the sync path stay separate.

The root page is now just the general idea, why it exists, what it gives you, and the choose-between-the-two table — down from ~380 lines originally to 87.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Added, and it reframed the whole section — this is the root cause rather than a symptom, so it now leads.

"Why a new runtime" opens with OPA's data store being a single in-memory JSON document, then draws the two consequences: no indexes and no concept of edges, so it cannot be optimized for graph traversal or for the relationship-heavy queries ReBAC and derived roles need; and the memory cost.

It also connects forward — the New Edge PDP moves those queries out of the JSON document into a store built for traversal, which OPA reaches over the query loopback while it evaluates policy. The Architecture page makes the same point against the request path.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Added as a blockquote with attribution and a link to [OPA: Policy Performance](https://www.openpolicyagent.org/docs/policy-performance). Much better than my hand-waving — it turns "memory scales with your data" into a number from an authoritative source, and 20x is a striking one.

The resource-footprint table on the How It Works page now carries the same citation, replacing the sentence that pointed at sharding.

dshoen619: Removed — all three references on these pages are gone (this one, the choose-between-the-two row, and a sentence in the resource-footprint section). The memory argument now rests on the OPA 20x citation instead of on sharding being the container PDP's answer.

On the wider cleanup: seven other files in this repo still reference sharding —

- `how-to/deploy/overview.mdx`
- `how-to/deploy/on-prem/management.mdx`
- `how-to/deploy/on-prem/reference.mdx`
- `how-to/deploy/on-prem/installation.mdx`
- `how-to/manage-data/loading-data.mdx`
- `concepts/pdp/overview.mdx` (has a "Sharded Edge PDPs" section)
- `concepts/pdp/configuration.mdx`

That is outside this PR's scope, but it is the set a cleanup ticket would need to cover — `concepts/pdp/overview.mdx` especially, since a whole section there presents sharding as a recommended deployment model.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Removed here too — that row now reads "Large data sets without a large memory footprint", which is the actual benefit rather than a comparison against a deprecated feature.

docs/concepts/pdp/new-edge-pdp.mdx
dshoen619: Flagging rather than actioning this one, because it reverses an explicit call by the PR author.

There was a configuration page earlier in this PR. After your note that some env vars are likely to change, the author asked for it to be removed rather than shipped with a caveat — first the variable tables, then the section entirely. So restoring it is their decision, not mine to quietly undo, and I have put your request to them.

Worth saying I think your framing resolves the original objection: a page marked beta and explicitly open to change is a different proposition from a reference that silently goes stale. If the author agrees, it comes back quickly — the content still exists in this branch's history at 054aaec^ and can be restored with the beta callout added.

Note the current pages do not leave readers with nothing: `PDP_API_KEY` is documented as the only required value, and the operational settings that matter (persistent storage, ≥40s `terminationGracePeriodSeconds`, probe targets and ports, memory sizing) are all on the Deployment page — just described rather than pinned to variable names.

dshoen619: Done — the configuration page is back, marked beta, and in the sidebar under **New Edge PDP → Configuration**.

Your framing is what made it work. It opens with a caution stating the surface is not stable, that variables may be renamed, replaced or removed before GA, that it should be treated as a working reference rather than a contract, and that readers should pin an image tag and re-read on upgrade — plus an ask to tell us if they are depending on a specific variable so we can flag it before it moves.

Restored from before the deletion and brought up to date: "container PDP" throughout, and deep links into pages that did not exist when it was first written (the port table on Architecture, health-and-readiness and deployment-requirements on Deployment, resource-footprint on How It Works). The storage-engine tuning advice now points here instead of at support.

Covers credentials, storage and ports, behavior and tuning, child-process log levels with the data-exposure warning, and the storage-engine knobs.

omer9564 (approved):
LGTM

